PR #35, opened 2022-05-06
Feature/gfi-json-display

JoshGx2000:
If the info format is "unavailable" for the layers when attempting to get the feature info, then the GetFeatureInfo request is skipped – a popup saying that layer data is currently unavailable is shown instead.

New popup also used in the event that the JSON popup has no data to display.

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -5,7 +5,7 @@
   (:require [clojure.string :as string]
             [cljs.spec.alpha :as s]
             [imas-seamap.utils :refer [encode-state ids->layers]]
-            [imas-seamap.map.utils :refer [applicable-layers layer-name bounds->str region-stats-habitat-layer wgs84->epsg3112 feature-info-html feature-info-json get-layers-info-format group-basemap-layers]]
+            [imas-seamap.map.utils :refer [applicable-layers layer-name bounds->str region-stats-habitat-layer wgs84->epsg3112 feature-info-html feature-info-json get-layers-info-format group-basemap-layers feature-info-none]]
             [ajax.core :as ajax]
             [imas-seamap.blueprint :as b]
             [reagent.core :as r]
@@ -143,11 +143,10 @@
 (defn got-feature-info [db [_ request-id priority point info-format response]]
   (if (not= request-id (get-in db [:feature-query :request-id]))
     db                           ; Ignore late responses to old clicks
-
     (let [feature-info (case info-format
                          "text/html" (feature-info-html response)
                          "application/json" (feature-info-json response)
-                         (feature-info-html response))
+                         (feature-info-none))
           db' (update-in db [:feature-query :response-remain] dec)
           {:keys [response-priority response-remain candidate had-insecure?]} (:feature-query db')
           higher-priority? (< priority response-priority)
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -25,71 +25,68 @@
         {:db db
          :put-hash (encode-state db)}))))
 
+(defn get-feature-info-request
+  [info-format request-id by-server size bounds x y point]
+  (let [layers->str   #(->> % (map layer-name) reverse (string/join ","))]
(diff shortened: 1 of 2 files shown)


frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -226,14 +234,16 @@
...
+        property-rows (str/join "" (map (fn [property] (property-to-row property)) properties))
+        properties []
+        id nil]
     (if (or id (not-empty properties))
       {:info (str
               "<div class=\"feature-info-json\">"
               "<h4>" id "</h4>"
               "<table>" property-rows "</table>"
               "</div>")}
-      {:status :feature-info/empty})))
```
markhepburn, line 236: This worries me a bit -- is this saying that there'll never be no results returned?  Even if you click in the ocean, or on land?

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -31,50 +31,61 @@
...
+                                   :response-priority 99
+                                   :had-insecure?     had-insecure?
+                                   :candidate         nil}
+                   :feature       {:status   :feature-info/waiting
+                                   :location point})}
+        
+        info-format
+        (assoc
+         :http-xhrio
+         (when info-format
```
markhepburn, line 53: This looks redundant -- isn't the cond-> clause "info-format" already implicitly a "truthy" condition?

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -31,50 +31,61 @@
         ;; Note, top layer, last in the list, must be first in our search string:
         layers->str   #(->> % (map layer-name) reverse (string/join ","))
         request-id    (gensym)
-        had-insecure? (->> db :map :active-layers (some #(is-insecure? (:server_url %))))]
+        had-insecure? (->> db :map :active-layers (some #(is-insecure? (:server_url %))))
+        info-format (get-layers-info-format active-layers)]
```
markhepburn, line 35: Minor, but might as well align these

frontend/src/cljs/imas_seamap/map/events.cljs
markhepburn: In fact looking further, I think this would be better extracted as a function at this point -- you're using cond-> to implement if-then-else, and the whole lot is getting a bit hard to follow.

markhepburn (changes requested):
Mind having a bit more of a tidy please?  There's a couple of things, both stylistic and possibly a logic / semantic issue (handling "normal" no results).

Also I'm about to push, so it may not merge cleanly... sorry :)

frontend/src/cljs/imas_seamap/map/utils.cljs
JoshGx2000: You're absolutely right. Reversing that change now.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: For your first comment, yep, that's silly me who forgot to delete some code I made when trying out solutions. For your second comment, I'll work on that now.

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -25,61 +25,68 @@
...
+      (let [params {:REQUEST       "GetFeatureInfo"
+                    :LAYERS        layers
+                    :QUERY_LAYERS  layers
+                    :WIDTH         (:x size)
+                    :HEIGHT        (:y size)
+                    :BBOX          (bounds->str bounds)
+                    :FEATURE_COUNT 5
+                    :STYLES        ""
+                    :X             x
+                    :Y             y
```
markhepburn, line 49: Final request please -- would you mind merging develop into your branch?  I made some concurrent edits that will be clobbered by these lines otherwise (essentially, x/y/width/height is fixed, and bounds is calculated separately from the viewport)

markhepburn (changes requested):
Looks good thanks -- there's a merge conflict you'll need to take care of first though (sorry), I'm suggesting merging develop to resolve it in this branch before we merge the PR.

PR #36, opened 2022-05-06
Feature/isa-90-actions-update-index

JoshGx2000:
Simple change; index.html is now in a setup that works for dev, and it's conditionally updated in the github actions (tested locally, so it should work).

### DIFF
--- a/.github/workflows/buildfrontend.yml
+++ b/.github/workflows/buildfrontend.yml
@@ -64,6 +64,14 @@ jobs:
       - name: Build project
         run: npx shadow-cljs release ${{ env.PROFILE }}
 
+      - name: Conditionally update app.js path in index (master)
+        if: endswith(github.ref, '/master')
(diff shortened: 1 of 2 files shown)


.github/workflows/buildfrontend.yml
```diff
@@ -64,6 +64,14 @@ jobs:
       - name: Build project
         run: npx shadow-cljs release ${{ env.PROFILE }}
 
+      - name: Conditionally update app.js path in index (master)
+        if: endswith(github.ref, '/master')
+        run: sed -i 's/app.js/compiled\/app\/app.js/' 'resources/public/index.html'
+
+      - name: Conditionally update app.js path in index (develop)
+        if: endswith(github.ref, '/master')
```
markhepburn, line 72: :)

(Other than that, looks good -- I forgot about the complication with master/develop.  I feel like there might be a better way around that, at the build level, but let's leave it for now.

markhepburn (changes requested):
Just a copy-paste error I think -- otherwise looks good thanks.

PR #40, opened 2022-05-11
Feature: (ISA-120) Drawer component

JoshGx2000:
A React component for a drawer has been created (https://blueprintjs.com/docs/#core/components/drawer).
The drawer is currently used in a basic "seamap drawer" component, which has its state (opened/closed) stored in the frontend db.
There are events to change the seamap drawer state, including a toggle hooked up to a hotkey ('a', which might need to be changed).

The seamap drawer is not currently in use, but the component can easily be added to the layout app in views.cljs

### DIFF
--- a/frontend/src/cljs/imas_seamap/views.cljs
+++ b/frontend/src/cljs/imas_seamap/views.cljs
@@ -748,6 +748,5 @@
       {:id "third-party-group" :helperText "Layers from other providers (eg CSIRO)"}]
      [welcome-dialogue]
      [info-card]
-     [loading-display]
-     [components/seamap-drawer]]))
+     [loading-display]]))
 
--- a/frontend/src/cljs/imas_seamap/views.cljs
+++ b/frontend/src/cljs/imas_seamap/views.cljs
@@ -693,6 +693,9 @@
        {:label "Toggle Sidebar"         :combo "s"}
        [:ui.sidebar/toggle])
       (keydown-wrapper
+       {:label "Toggle Seamap Drawer"   :combo "a"} ;; TODO: Better label and perhaps different key mapping?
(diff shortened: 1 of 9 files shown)


frontend/src/stories/Drawer/Drawer.jsx
```diff
@@ -0,0 +1,43 @@
...
+				case "LEFT":
+				case "Left":
+				case "left":
+				case "l":
+					return BPCore.Position.LEFT;
+				default:
+					return BPCore.Position.RIGHT;
+			}
+		}
+	)(position);
```
olivergeorge, line 28: I haven't bothered with this type of thing.  Feel unnecessary and encourages sloppy values.

I suspect BPCore.Position.TOP has a concrete string value "top" or whatever.  Let's just use that.

JoshGx2000: Yes, they do. I hadn't realised. Change just added.

I also realise that I neglected to add prop types, so those have been added too.

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -692,6 +692,9 @@
       (keydown-wrapper
        {:label "Toggle Sidebar"         :combo "s"}
        [:ui.sidebar/toggle])
+      (keydown-wrapper
+       {:label "Toggle Seamap Drawer"   :combo "a"} ;; TODO: Better label and perhaps different key mapping?
```
markhepburn, line 696: Since you raised it -- "seamap" probably doesn't need to be included :)  ("d"is available, but I suspect we'll overhaul these with the new UI anyway so I wouldn't waste any time over it)

markhepburn:
lgtm

PR #41, opened 2022-05-16
Feature: (ISA-120) Drawer Content

JoshGx2000:
New features:
• PanelStack React component and interop – initially a bit fiddly for displaying panels, but the code should be quite neat now.
• Drawer clojure component extracted from Seamap drawer, and Seamap drawer moved to views.
• Seamap drawer content additions:
	◦ Heading replaced with Seamap logo image (obtained from: https://seamapaustralia.org/wp-content/uploads/2017/10/Seamap2_V2_RGB.png) – do we need to ask permission to use this from them?
	◦ Main body replaced with PanelStack filled with some basic content, such as the control buttons (non-functional), access to the catalogue layer panels (WIP, filled with placeholders), and settings (non-functional).
• Seamap drawer PanelStack state stored in frontend DB, with new events and a sub for changing and reading the state of which panels are active.


frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -614,6 +614,110 @@
...
+   :content "Third-Party Layers (WIP)"})
+
+(def seamap-drawer-panels
+  {:drawer-panel/habitat-layers    habitat-layers-panel
+   :drawer-panel/bathy-layers      bathy-layers-panel
+   :drawer-panel/imagery-layers    imagery-layers-panel
+   :drawer-panel/management-layers management-layers-panel
+   :drawer-panel/thirdparty-layers thirdparty-layers-panel})
+
+(defn seamap-drawer
```
markhepburn, line 697: Just a general naming thing; I'm not keen on the "seamap-" prefix, it's kind of superfluous in the seamap codebase.

frontend/src/stories/PanelStack/PanelStack.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
markhepburn, line 1: Another minor thing -- not for this PR -- can we rename the "stories" directory to "ui" or something more meaningful?

frontend/src/stories/PanelStack/PanelStack.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from "prop-types";
+import * as BPCore from "@blueprintjs/core";
+
+export function Panel({content}) {
+	return content
+}
```
markhepburn, line 7: Huh!  I don't know that I have a comment, I'm just curious: is that idiomatic?

frontend/src/stories/PanelStack/PanelStack.stories.jsx
```diff
@@ -0,0 +1,89 @@
...
+			panels={panels}
+			onClose={close}
+			onOpen={open}
+		/>
+	);
+};
+
+export const SimplePanelStack = FieldTemplate.bind({});
+SimplePanelStack.args = {
+};
```
markhepburn, line 89: I'm really being pedantic here, but see if your editor has a setting to always ensure a trailing newline (it's more or less a unix convention; it doesn't break anything I can think of, but that's why github is adding that symbol)

frontend/src/stories/PanelStack/PanelStack.css
```diff
@@ -0,0 +1,14 @@
+.bp3-panel-stack {
+	height: 200px;
+}
+
+.panel-stack-content {
+	padding: 8px;
+	display: flex;
+    align-items: center;
```
markhepburn, line 8: I don't know if I'm overly pedantic or this is a super-subtle joke, but align... isn't aligned :)

frontend/src/cljs/imas_seamap/events.cljs
```diff
@@ -470,4 +470,10 @@
   (assoc-in db [:display :seamap-drawer] true))
 
 (defn seamap-drawer-close [db _]
-  (assoc-in db [:display :seamap-drawer] false))
\ No newline at end of file
+  (assoc-in db [:display :seamap-drawer] false))
+
+(defn drawer-panels-open [db [_ panel props]]
+  (update-in db [:display :drawer-panels] #(conj % {:panel panel :props props})))
```
olivergeorge, line 476: (update-in db [:display :drawer-panels] conj {:panel panel :props props}))

frontend/src/cljs/imas_seamap/db.cljs
```diff
@@ -45,6 +45,7 @@
    :display         {:help-overlay    false
                      :welcome-overlay false
                      :seamap-drawer   false
+                     :drawer-panels   []
```
markhepburn, line 48: Let's get this added to the spec (also the drawer, noting other feedback about naming too).
Also worth thinking about how we communicate how this list is used (is the current panel first or last in the last?  I see it's "last" from the events below, but eg calling it drawer-panel-stack, and I think there's push/pop functions even which might be clearer?)

markhepburn (commented):
I've made mostly stylistic comments; it would good to have @olivergeorge feedback too, perhaps more on the jsx side. (and I mentioned in person that from an "atomic" perspective it might be a more self-contained or focussed PR with just the panel-stack, rather than panel-stack + usage.  On the other hand, it's handy seeing how it fits into the UI, so I'm not complaining, just a general thought).  Looking good otherwise.

frontend/src/cljs/imas_seamap/db.cljs
markhepburn: I was half right: there's a [pop](https://clojuredocs.org/clojure.core/pop), but no push!  I guess that's what conj does.  @olivergeorge can comment.

frontend/src/stories/PanelStack/PanelStack.stories.jsx
JoshGx2000: Done!

frontend/src/stories/PanelStack/PanelStack.css
JoshGx2000: Oh, that's completely accidental XD

frontend/src/cljs/imas_seamap/views.cljs
JoshGx2000: Hmm, yeah. Just calling it "drawer" is a little generic, and might get it confused with the generic component (though maybe not, as one is in the components file and the other is in the views file now. Maybe " sidebar-drawer" works; do you have thoughts on this?

frontend/src/stories/PanelStack/PanelStack.jsx
JoshGx2000: Done!

frontend/src/cljs/imas_seamap/events.cljs
JoshGx2000: Thanks, I didn't realise that would work 👍

frontend/src/cljs/imas_seamap/db.cljs
JoshGx2000: Okay, so I've renamed the events to push and pop, and drawer-panels to drawer-panel-stack to better reflect their nature as a stack.

frontend/src/stories/PanelStack/PanelStack.jsx
JoshGx2000: Technically I am extracting content from the props, but I've now simplified it to an arrow function in the PanelStack, which I hadn't thought to do initially.

PR #42, opened 2022-05-16
Feature: (ISA-120) Drawer map layers panels

JoshGx2000:
Just some simple changes that add functionality to the drawer.
Only one hack that I'm concerned with and want someone to take a look at: for the subs in "child" panels to update, there needs to be an unnecessary sub in "base-panel" (line 635).

### DIFF
--- a/frontend/src/cljs/imas_seamap/views.cljs
+++ b/frontend/src/cljs/imas_seamap/views.cljs
@@ -930,6 +930,5 @@
       {:id "third-party-group" :helperText "Layers from other providers (eg CSIRO)"}]
      [welcome-dialogue]
      [info-card]
-     [loading-display]
-     [seamap-drawer]]))
+     [loading-display]]))
 
--- a/frontend/src/cljs/imas_seamap/views.cljs
+++ b/frontend/src/cljs/imas_seamap/views.cljs
@@ -616,56 +616,62 @@
 
 (defn base-panel
   []
-  {:title   "Base Panel"
-   :content [:div
-             [:div.seamap-drawer-group
-              [:h1.bp3-heading.bp3-icon-settings
-               "Controls"]
-              [b/button
-               {:icon     "edit"
-                :text     "Draw Transect"}]
-              [b/button
-               {:icon     "widget"
-                :text     "Select Region"}]
-              [b/button
-               {}
-               [:span
-                [:i.fa.fa-magic]
-                "Enable Automatic layer Selection"]]]
-             [:div.seamap-drawer-group
-              [:h1.bp3-heading.bp3-icon-list-detail-view
-               "Catalogue Layers"]
-              [b/button
-               {:icon     "home"
-                :text     "Habitat Layers"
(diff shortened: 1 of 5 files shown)


frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -714,6 +786,7 @@
       :isOpen   open?
       :onClose  #(re-frame/dispatch [:seamap-drawer/close])}
      [:div.seamap-drawer
+      {:key 1}
```
markhepburn, line 789: I'm sure this is a todo, but noting here just in case (it's probably fine in this instance, it just looks like a code smell!  If these are panels, the name is probably fine as an id)

markhepburn (changes requested):

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -616,76 +632,132 @@
...
-               {:icon     "more"
-                :text     "Third-Party Layers"
-                :on-click #(re-frame/dispatch [:drawer-panel-stack/push :drawer-panel/thirdparty-layers])}]]
-             [:div.seamap-drawer-group
-              [:h1.bp3-heading.bp3-icon-cog
-               "Settings"]
-              [b/button
-               {:icon     "undo"
-                :text     "Reset Interface"}]]]})
+  (let [_ @(re-frame/subscribe [:map/layers])] ; Subs for "child" panels won't update without this sub (which shouldn't be affecting them?)
```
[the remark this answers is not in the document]
JoshGx2000: Yeah, it'd be nice to go through this, because it greatly confuses me, and I'm sure I'm doing something wrong.

frontend/src/cljs/imas_seamap/views.cljs
JoshGx2000: Yeah, I'm not sure what's up with this, but it doesn't appear to be for the panels (happens for a drawer without the panels, so it's potentially related to the drawer?

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -614,96 +630,161 @@
...
+      :expanded-fn   expanded-layers
+      :opacity-fn    layer-opacities}]]})
+
+(defn drawer-panel-selection
+  [panel map-layers region-stats]
+  (let [{:keys [panel props]} panel
+        {:keys [groups active-layers loading-layers error-layers expanded-layers layer-opacities]} map-layers
+        {:keys [habitat-layer]} region-stats]
+    (case panel
+      :drawer-panel/layer-panel
```
markhepburn, line 740: This is fine, really.  I think I might have used a defmulti (dispatching on panel) elsewhere to do similar stuff which is worth considering.

markhepburn (approved):
(without looking at the code in its entirety) that's looking a lot more idiomatic and sane now, nice work.

frontend/src/cljs/imas_seamap/views.cljs
JoshGx2000: I might consider doing this later, considering I haven't really used defmulti at all yet.

PR #44, opened 2022-05-20
Feature: (ISA-88) GFI Multiserver Logic

JoshGx2000:
Fixes https://jira.its.utas.edu.au/browse/ISA-88

Ensures that results returned from all servers are then combined into the single popup that is displayed.

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -150,7 +150,7 @@
 (defn toggle-ignore-click [db _]
   (update-in db [:map :controls :ignore-click] not))
 
-(defn responses-feature-info [responses]
+(defn responses-feature-info [db point]
   (letfn [(response-to-info ; Converts a response and info format into readable information for the feature info popup
             [response]
             (if-let [{:keys [response info-format]} response] ; If we have a response; then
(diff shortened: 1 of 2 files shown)


frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -89,7 +83,7 @@
         img-bounds    (bounds-for-zoom point size bounds img-size)
         ;; Note, top layer, last in the list, must be first in our search string:
         request-id    (gensym)
-          had-insecure? (->> db :map :active-layers (some #(is-insecure? (:server_url %))))
+        had-insecure? (->> db :map :active-layers (some #(is-insecure? (:server_url %))))
```
markhepburn, line 86: Ignore for now, but just for context -- we might be able to finally remove this at some point.  The gist is, browsers don't like making http ajax requests from a https page (or a bit more confusingly for me, even https from an http page), so we had this workaround because there was always some rogue GA or CSIRO layer still on http.  That _may_ finally be at an end.  (Like I said, leave it in, I'm just talking out loud)

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -159,32 +150,38 @@
...
-        (assoc :feature (if higher-priority? candidate' candidate))))))
-
-(defn got-feature-info-error [db [_ _request_id priority _]]
-  (let [db (update-in db [:feature-query :response-remain] dec)
-        {:keys [response-priority response-remain]} (:feature-query db)
-        higher-priority? (< priority response-priority)]
-    (if (and (zero? response-remain) higher-priority?)
-      (assoc-in db [:feature :status] :feature-info/error)
-      db)))
+          feature-info (update feature-info :info str (:info candidate)) ;; Merge existing candidate info into retrieved info
```
markhepburn, line 163: I think I'm all on-board apart from this line.  It _probably_ works, but just concatenating responses (assuming I'm not mis-reading, and it's just mashing strings together?) seems messy, and error-prone at some point (actually, what happens if you get a mix of html + json?).  I think better would be to collect a list of responses, and then we have more flexibility in rendering (it might be a big scrolling popup, it might be tabs with one response each, etc).

markhepburn (changes requested):
Happy to talk that through, particularly if I'm mis-reading or have missed the point.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: I've removed the single "candidate" key from the query, and just store each server response in a list until all of them arrive, where they are then combined.
Doing this also allowed me to neaten up a lot of the code in got-feature-info and got-feature-info-error.
Happy for more feedback.

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -159,32 +150,54 @@
...
+        response-infos (map
+                        (fn [response]
+                          (if response
+                            (let [{:keys [response info-format]} response]
+                              (case info-format
+                                "text/html" (feature-info-html response)
+                                "application/json" (feature-info-json response)
+                                (feature-info-none)))
+                            {:status :feature-info/empty}))
+                        responses)
```
markhepburn, line 164: So, a few things here:

- My eyes jumped to this block of code (and response-info below); it's quite blocky and not super readable.  Any time you have an anonymous function in a map or (particularly) reduce, if it's more than a couple of lines it attracts attention I find.
- https://clojuredocs.org/clojure.core/if-let :)
- In general I think I'd write functions to at least name what you're doing (https://clojuredocs.org/clojure.core/letfn might be useful too)

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -159,32 +150,54 @@
...
+        response-info  (reduce
+                        (fn [acc response-info]
+                          (cond-> acc
+                            (:info response-info)
+                            (update :info str (:info response-info))
+                            (:status response-info)
+                            (assoc :status (:status response-info))
+                            (or (:info response-info) (:info acc))
+                            (dissoc :status)))
```
markhepburn, line 173: Case in point: what's this doing, exactly?  From what I can see, you're updating a single dict with :info/:status, over a collection (which seems like it would be order-dependent, and perhaps that's the point).  It's probably correct, but it's hard to follow and I'd say at least needs explanation (starting with a name as suggested above, and I'd also suggest comments for each clause)

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -207,7 +220,8 @@
       ;; If category is contours, turn it into bathymetry but with a contours attribute set:
       (as-> lyr (if (= (:category lyr) :contours) (assoc lyr :contours true) lyr))
       (update :category    #(if (= % :contours) :bathymetry %))
-      (update :server_type (comp keyword string/lower-case))))
+      (update :server_type (comp keyword string/lower-case))
+      (update :server_url #(if (= % "https://geoserver-dev.imas.utas.edu.au/geoserver/wms") "https://geoserver.imas.utas.edu.au/geoserver/wms" %))))
```
markhepburn, line 224: What's going on here; is this debug code that snuck through?

markhepburn (changes requested):
The general idea is sound, but I think there's some decent clarity and code style issues that we can address before finishing off.  Have a go at clarifying that main processing block I've commented on, and we can have another look.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Whoops, yes!

JoshGx2000: Well, that's been removed immediately.

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -159,32 +150,51 @@
...
+
+(defn got-feature-info-error [db [_ request-id point _]]
+  (if (not= request-id (get-in db [:feature-query :request-id]))
+    db ; Ignore late responses to old clicks
+    (-> db
+        (update-in [:feature-query :response-remain] dec)
+        (update-in [:feature-query :responses] conj nil)
+        (cond->
+         (zero? (dec (get-in db [:feature-query :response-remain])))
+          (display-feature-info point))))) ;; If this is the last response expected, update the displayed feature
```
markhepburn, line 197: This is the only bit that looks a bit suss now; I think it's just because I'm used to seeing `update-in` and the like in these chains, where you're explicit about what's being updated, whereas this is a bit opaque.  I don't immediately see a better way though.  Perhaps it's the name?  (I guess you're not really displaying it here, you're formatting for display)

(Also, is that an indenting issue I just noticed?)

markhepburn: Hmm, there's a single update at the end; perhaps it could instead return the formatted response, and got-feature-info assocs that on instead?  That feels like a better separation of concerns to me.

markhepburn (changes requested):
Looks good, I just commented on one final bit, which is largely stylistic but still important I think.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Sounds good

JoshGx2000: I've restructured the code to just return the summarised response, as suggested.

markhepburn (approved):

PR #45, opened 2022-05-24
Feature: (ISA-23) Serialized hashstate

JoshGx2000:
The current changes allow one to enter a shorthand hashstate (in the form `#view=<shortcode>`) to load a hashstate stored in the database. There is no infrastructure for adding new hashstates.


backend/catalogue/models.py
```diff
@@ -138,3 +138,12 @@ class BaseLayerGroup(models.Model):
 
     def __str__(self):
         return self.name
+
+
+@python_2_unicode_compatible
+class SerializedHashstate(models.Model):
+    id = models.CharField(max_length = 36, primary_key=True)
```
markhepburn, line 145: Django does actually have a UUID field from memory (can't offhand remember if sqlserver has native support though...and obviously only useful if we use uuids anyway)

backend/webapp/urls.py
```diff
@@ -19,6 +19,7 @@
 router.register(r'priorities', GroupPriorityViewset)
 router.register(r'descriptors', DescriptorViewset)
 router.register(r'baselayergroups', BaseLayerGroupViewset)
+router.register(r'serializedhashstates', SerializedHashstateViewset)
```
markhepburn, line 22: I'm in favour of renaming that; it's quite "literal".  Even "savedstate" is probably more explanatory.

frontend/src/cljs/imas_seamap/core.cljs
```diff
@@ -60,6 +60,8 @@
     :re-boot                              events/re-boot
     :ajax/default-success-handler         (fn [db [_ arg]] (js/console.log arg) db)
     :ajax/default-err-handler             (fn [db [_ arg]] (js/console.error arg) db)
+    :get-serialized-hashstate             [events/get-serialized-hashstate]
+    :got-serialized-hashstate             events/got-serialized-hashstate
```
markhepburn, line 64: Ditto here.  "Serialised" isn't used anywhere else (it's "put-hash", etc).  It's literally true, but it's long and probably doesn't convey much more.

frontend/src/cljs/imas_seamap/events.cljs
```diff
@@ -79,6 +82,33 @@
...
+      shortcode
+      (assoc
+       :http-xhrio
+       [{:method :get
+         :uri             serialized-hashstate-url
+         :response-format (ajax/json-response-format {:keywords? true})
+         :on-success      [:got-serialized-hashstate]
+         :on-failure      [:ajax/default-err-handler]}])
+      (not shortcode)
+      (assoc :dispatch [:got-serialized-hashstate]))))
```
markhepburn, line 99: I haven't tried to follow this exactly, but two impressions:

- (minor stylistic) this looks like another cond-masquerading-as-an-"if" to me
- I might be reading this wrong, but I'm quite leery of triggering the :got-... event when you didn't request it (perhaps it's a naming thing).  Without looking into how tricky, initial load has either no existing state / it has an existing saved state / it has a short-code that we need to retrieve.  They're all separate execution paths, even if they wind up in the same place.  (the :seen-any-of clause for async-flow looks useful)

markhepburn: (in fact looking at the got-serialised-hashstate below, and really a lot of this code, it would be simplified if each case was a separate path; there's a lot of "cond->" followed by cases for whether there's a short code or not, etc, which just makes it hard to follow.

markhepburn (changes requested):
I think the two overriding impressions are:

- I'm not like the "serializedhashstate" name at all.  It's basically passable as a django model, less so as a django URL, and complicates the reading of it in the cljs code.  I think "saved-state" or something that reflects the actual intent (and is easier to read) is better, or at worst just hash-state (note, the existing code uses hash-state at the boundary, because that's literally what it is -- it's the hash-string in the browser's url bar.  You're using it as a core concept in the code, where I think it's losing its connection to what it's referring to).
- In general it's hard to follow, because I think yo'ure finding that lots of places need to differentiate between having the state or not  (this can easily happen when you add a quick change the first time you need to... then you're kind of stuck and have to keep making the same adjustment!).  Breaking those up into separate execution paths will make things much easier to follow I think.

frontend/src/cljs/imas_seamap/events.cljs
JoshGx2000: I agree. I'll rewrite those then 👍

markhepburn (approved):
This looks good now -- anything else needed, or can that be merged?

(I thought it was already done, to be honest)

PR #46, opened 2022-05-26
Feature: (ISA-124) Zoom level info

JoshGx2000:
Just want to check that I added the leaflet zoom info source code to the correct place. Everything runs, but I want to know that I followed our best practices, etc.


frontend/shadow-cljs.edn
```diff
@@ -1,4 +1,4 @@
-{:source-paths ["src/cljs" "src/gen"]
+{:source-paths ["src/cljs" "src/gen" "src/leaflet_zoominfo"]
```
markhepburn, line 1: I think I'd change this to src/thirdparty ("vendor" is also common)

markhepburn (commented):
Just the one bit of minor feedback

frontend/shadow-cljs.edn
JoshGx2000: Cool, I'll make this change and then merge.

PR #48, opened 2022-05-27
Feature: (ISA-136) Google Maps pills

JoshGx2000:
Google Saps style pill buttons added over top of map. Placeholder button functionality used until we know what we actually want to use them for.

Quite CSS heavy, but I think I accounted for all the sidebar states accurately. Got to remember to account for the different absolute positioning when we decide we want it to move with the drawer rather than the Leaflet panel.

### DIFF
--- a/frontend/resources/public/css/app.css
+++ b/frontend/resources/public/css/app.css
@@ -860,7 +860,7 @@ input.leaflet-control-layers-selector[type="radio"] {
 }
 /* L.Control.Zoominfo.css end */
 
-.floating-pill-buttons {
+.floating-pills {
     position: absolute;
     z-index: 2000;
     top: -40px;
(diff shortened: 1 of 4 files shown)


frontend/resources/public/css/app.css
```diff
@@ -859,3 +859,62 @@ input.leaflet-control-layers-selector[type="radio"] {
...
+
+@media (min-width:768px) {
+	.floating-pills {
+        top: 10px;
+	}
+}
+
+@media (min-width:768px) and (max-width:991px) {
+	.floating-pills {
+		left: calc(305px + 20px);
```
markhepburn, line 880: My only question is where these numbers are coming from; is that 305px etc from the current sidebar?

JoshGx2000: Yes, that's right

JoshGx2000: 305px (etc.) is the sidebar width, and 20px is the 10px of "padding" either side of the sidebar.

markhepburn: Cool; something to keep in mind I guess (actually, is that what the sidebar does to the rest of the floating controls?  I assume so)..

PR #49, opened 2022-05-30
Feature: (ISA-135) Arbitrary layer categories

JoshGx2000:
Designed to solve ISA-135 (https://jira.its.utas.edu.au/browse/ISA-135).

The different layer tabs/groups now all display their layers in layer catalogues (example, the habitat layers below).
Screen Shot 2022-05-30 at 4 34 59 pm

There have been a few changes to the app state to support this, involving independent node states (expanded/collapsed) per tab. Some parts of the catalogue layers view have been updated, allowing for different tabs and category groupings per layer group (though this remains unused, and all layer groups use "By Organisation" and "By Category" categorisation).


frontend/src/cljs/imas_seamap/db.cljs
```diff
@@ -47,8 +47,16 @@
...
+                     :catalogue          {:habitat     {:tab "org"     ; NOTE: Move away from map to vector to support arbitrary number of tabs/groups?
+                                                        :expanded #{}}
+                                          :boundaries  {:tab "org"
+                                                        :expanded #{}}
+                                          :bathymetry  {:tab "org"
+                                                        :expanded #{}}
+                                          :imagery     {:tab "org"
+                                                        :expanded #{}}
+                                          :third-party {:tab "org"
+                                                        :expanded #{}}}
```
markhepburn, line 59: This is the obvious bit that stands out -- remember we'll have an arbitrary number of categories soon, not just this 5.  It's fine if you're remembering that and just want to do one thing at a time (I encourage that, even!), but do keep it in mind.
This is probably more relevant below where you've updated the app-state spec (thank you!); this is good, but you won't be able do that shortly, so it might be worth just making this section a s/map-of or whatever the syntax is, instead of speccing every key.

markhepburn (commented):
LGTM, noting the comment about hard-coding categories.

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -646,23 +649,68 @@
...
       {:icon     "more"
        :text     "Third-Party Layers"
-       :on-click #(re-frame/dispatch [:drawer-panel-stack/push :drawer-panel/thirdparty-layers])}]]
+       :on-click #(re-frame/dispatch [:drawer-panel-stack/push :drawer-panel/catalogue-layers
+                                      {:group :third-party
+                                       :title "Third-Party Layers"
+                                       :tabs
+                                       [{:id         "org"
+                                         :title      "By Organisation"
+                                         :categories [:organisation :data_classification]}
```
markhepburn, line 710: Small thing, but is it feasible to make :data_classification into :data-classification?  (it depends on where it comes from, obviously)

markhepburn (approved):
Comment added (minor stylistic thing), but lgtm otherwise.

frontend/src/cljs/imas_seamap/views.cljs
JoshGx2000: It comes from the model:
Screen Shot 2022-05-31 at 5 31 46 pm
which in turn comes from the DB:
Screen Shot 2022-05-31 at 5 32 10 pm

So we _could_ change it by massaging the data, but I'm not sure if we want to?

PR #50, opened 2022-06-01
Feature: (ISA-1) transect layer names

JoshGx2000:
Updated the transect query to include the layer name of the polygon intersected by each segment. Updates to both `HabitatTableType.sql` and `path_intersections.sql` that have been applied to the dev database.

Transect tooltip now includes layer name for each segment.


markhepburn (approved):
LGTM -- not as scary as you imagined? :)

PR #53, opened 2022-06-02
Feature: (ISA-144) Removed additional sidebar tabs

JoshGx2000:
Simple change: removed all the additional sidebar tabs.

The tab enclosing the active layers tab content hasn't been removed because it would destroy the current styling and also leave the sidebar unable to be collapsed:
Screen Shot 2022-06-02 at 10 35 34 am

If fixing this is seen as worth it, then I can work on replicating the style, but for now I think what we have will do (especially when we don't even know for certain what we want the sidebar to look like).
Screen Shot 2022-06-02 at 12 38 09 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/views.cljs
+++ b/frontend/src/cljs/imas_seamap/views.cljs
@@ -837,7 +837,7 @@
       :hasBackdrop false}
      "With cooler content ;-)"]))
 
-(defn seamap-sidebar []
+(defn active-layers-sidebar []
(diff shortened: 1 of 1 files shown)


markhepburn (approved):
Thanks for the screenshots -- it looks a bit weird without any title bar, but given this is temporary I'm ok with it.

JoshGx2000:
> Thanks for the screenshots -- it looks a bit weird without any title bar, but given this is temporary I'm ok with it.

Actually, I'm bad at wording, because the one I ended up choosing was the one _with_ the title bar.

PR #59, opened 2022-06-06
Feature: (ISA-4) Automatically centre popup

JoshGx2000:
Automatically centering popup when GFI request is made
Screen Shot 2022-06-06 at 1 33 53 pm
Screen Shot 2022-06-06 at 1 33 59 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -45,6 +45,18 @@
      :east (+ lng (/ img-x-bounds 2))
      :west (- lng (/ img-x-bounds 2))}))
 
+(defn bounds-for-point
+  "Uses current bounds and a map point coordinate to get the map bounds centered
+   on that point."
+  [{:keys [lat lng]}
+   {:keys [north south east west]}]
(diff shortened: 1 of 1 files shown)


frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -45,6 +45,18 @@
      :east (+ lng (/ img-x-bounds 2))
      :west (- lng (/ img-x-bounds 2))}))
 
+(defn bounds-for-point
+  "Uses current bounds and a map point coordinate to get the map bounds centered
+   on that point."
+  [{:keys [lat lng]}
+   {:keys [north south east west]}]
+  (let [img-x-bounds (Math/abs (- west east))
```
markhepburn, line 53: I'm not super fussed, but "img" tends to have connotations of pixels rather than coordinates, in our context.

markhepburn: (cf getfeatureinfo, where you have both geographic bounds and x/y pixel coords)

markhepburn (approved):
I'm guessing this jumps a bit?  Anyway, let's deploy and see how it feels.

PR #60, opened 2022-06-07
Feature: (ISA-135) Remove category specific logic (catalogue panels)

JoshGx2000:
https://jira.its.utas.edu.au/browse/ISA-135

Adds category info to the DB, and uses new category display name field for the catalogue buttons and panels text.

### DIFF
--- a/backend/catalogue/viewsets.py
+++ b/backend/catalogue/viewsets.py
@@ -3,8 +3,8 @@
 # Released under the Affero General Public Licence (AGPL) v3.  See LICENSE file for details.
 from django.db.models import Value
 from django.db.models.functions import Coalesce
-from catalogue.models import Organisation, HabitatDescriptor, BaseLayer, Layer, LayerGroup, LayerGroupPriority, DataClassification, BaseLayerGroup, SaveState
-from catalogue.serializers import ClassificationSerializer, OrganisationSerializer, HabitatSerializer, BaseLayerSerializer, LayerSerializer, GroupSerializer, GroupPrioritySerializer, BaseLayerGroupSerializer, SaveStateSerializer
+from catalogue.models import Organisation, HabitatDescriptor, BaseLayer, Layer, LayerGroup, LayerGroupPriority, DataClassification, BaseLayerGroup, Category
+from catalogue.serializers import ClassificationSerializer, OrganisationSerializer, HabitatSerializer, BaseLayerSerializer, LayerSerializer, GroupSerializer, GroupPrioritySerializer, BaseLayerGroupSerializer, CategorySerializer
 from rest_framework import viewsets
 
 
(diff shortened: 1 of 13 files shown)


backend/catalogue/viewsets.py
```diff
@@ -3,8 +3,8 @@
 # Released under the Affero General Public Licence (AGPL) v3.  See LICENSE file for details.
 from django.db.models import Value
 from django.db.models.functions import Coalesce
-from catalogue.models import Organisation, HabitatDescriptor, BaseLayer, Layer, LayerGroup, LayerGroupPriority, DataClassification, BaseLayerGroup, SaveState
```
markhepburn, line 6: These are getting quite long -- let's clean up while we're at it ("import ..as models", then refer as models.BaseLayer, etc)

backend/webapp/urls.py
```diff
@@ -6,7 +6,7 @@
 from django.urls import include, re_path
 from rest_framework.routers import DefaultRouter
 
-from catalogue.viewsets import ClassificationViewset, DescriptorViewset, BaseLayerViewset, LayerViewset, GroupViewset, GroupPriorityViewset, OrganisationViewset, BaseLayerGroupViewset
```
markhepburn, line 9: Similar qualified import (import-as) could be used here.  (I know this is my fault that I'm asking you to clean up!)

frontend/src/cljs/imas_seamap/map/subs.cljs
```diff
@@ -65,6 +65,13 @@
 (defn map-base-layers [{:keys [map]} _]
   (select-keys map [:grouped-base-layers :active-base-layer]))
 
+(defn categories-map [db _]
+  (let [categories (get-in db [:map :categories])]
+    (reduce
+     (fn [categories {:keys [name] :as category}]
```
markhepburn, line 71: Second time in this PR I've seen this pattern (I always zoom in on uses of reduce, it's a prime candidate for "actually there's an existing library fn that does this").  How about extracting a util to index a collection by a key?  (If there isn't one already).  Have a bit of a play; one option is just to extract the categories case, another might be to pull out the map-key function, eg (into-map :name categories) or something like that (not claiming that's the best name)

backend/catalogue/viewsets.py
JoshGx2000: Good idea

frontend/src/cljs/imas_seamap/map/subs.cljs
JoshGx2000: I've extracted the function, as after some searching I've found examples of similar problems, which are not solved by a library function (e.g. https://stackoverflow.com/questions/47082635/map-from-list-of-maps).

backend/catalogue/viewsets.py
JoshGx2000: Imports cleaned-up

backend/webapp/urls.py
JoshGx2000: Imports cleaned-up

PR #62, opened 2022-06-09
Feature: (ISA-154 & ISA-156) Floating menu

JoshGx2000:
https://jira.its.utas.edu.au/browse/ISA-154
https://jira.its.utas.edu.au/browse/ISA-156

Lots of changes (maybe so many that it should have all been split).

Now using Sass for most styling. Lots of css from `app.css` was separated into different files focussing on certain elements of the page styling.

Leaflet panel removed and replaced with custom floating panel element which:
 • displays the active layers (ISA-154);
 • allows access to the omnibar for selecting layers (ISA-156); and
 • allows access to the catalogue drawer.

A few (small) style changes have been made to the floating pill buttons.

Screen Shot 2022-06-09 at 1 58 01 pm

Things to do:
 • Move top-left leaflet controls when no active layers?
 • Move over leaflet controls and pill buttons when catalogue drawer opens.


frontend/resources/public/index.html
```diff
@@ -10,6 +10,7 @@
     <link rel="stylesheet" href="https://unpkg.com/leaflet-draw@1.0.4/dist/leaflet.draw.css" />
     <link rel="stylesheet" href="https://unpkg.com/leaflet-sidebar-v2@1.0.0/css/leaflet-sidebar.min.css" />
     <link rel="stylesheet" href="css/app.css" />
```
markhepburn, line 12: Should that go, if we're not generating it?

markhepburn (approved):
I'll merge and we can see how it plays out; there's a small comment in there though, and a couple of minor things mentioned via chat.

PR #66, opened 2022-06-14
Feature: (ISA-160) Networks, parks, and zone support

JoshGx2000:
Networks, parks, and zones are retrieved from the DB (view each for networks, parks, zones (categories), and zones (IUCN)).

Some tweaks to select and omnibar field workings.

Basic UI set up for selecting the networks, parks, and zones:
Screen Shot 2022-06-14 at 12 48 43 pm


backend/habitat/viewsets.py
```diff
@@ -88,6 +88,11 @@
 PRJ_3112 = """PROJCS["GDA94_Geoscience_Australia_Lambert",GEOGCS["GCS_GDA_1994",DATUM["D_GDA_1994",SPHEROID["GRS_1980",6378137,298.257222101]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]],PROJECTION["Lambert_Conformal_Conic"],PARAMETER["standard_parallel_1",-18],PARAMETER["standard_parallel_2",-36],PARAMETER["latitude_of_origin",0],PARAMETER["central_meridian",134],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["Meter",1]]"""
 
 
+SQL_GET_NETWORKS = "select name from SeamapAus_Networks_View;"
+SQL_GET_PARKS = "select name, network from SeamapAus_Parks_View;"
+SQL_GET_ZONES = "select name from SeamapAus_Zones_View;"
+SQL_GET_ZONES_IUCN = "select name from SeamapAus_ZonesIUCN_View;"
```
markhepburn, line 94: Are these views ones that you've created?

markhepburn (approved):
There are a couple of little comments (mainly, I suspect there's some views that need to be checked in)

backend/habitat/viewsets.py
JoshGx2000: Yes, that's correct.

PR #67, opened 2022-06-15
Feature: (ISA-161) Query statistics for habitat

JoshGx2000:

Can now get and display the habitat statistics for a network or park. Statistics are displayed in a table on the state of knowledge sidebar, like so:
Screen Shot 2022-06-15 at 11 42 52 am

Currently does not work for zones. Does not display bathymetry or habitat observations information.

### DIFF
--- a/backend/habitat/viewsets.py
+++ b/backend/habitat/viewsets.py
@@ -520,13 +520,14 @@ def habitat_statistics(request):
     zone_iucn = request.query_params.get('zone-iucn')
 
     habitat_stats = []
+    total_area = 0
 
     with connections['transects'].cursor() as cursor:
         if (park):
             cursor.execute(SQL_GET_PARK_AREA, [park])
-            area = cursor.fetchone()[0]
+            total_area = cursor.fetchone()[0]
 
-            cursor.execute(SQL_GET_PARK_HABITAT_STATS, [area, park])
+            cursor.execute(SQL_GET_PARK_HABITAT_STATS, [total_area, park])
 
             while True:
                 try:
@@ -540,9 +541,9 @@ def habitat_statistics(request):
                         break
         elif (network):
             cursor.execute(SQL_GET_NETWORK_AREA, [network])
-            area = cursor.fetchone()[0]
+            total_area = cursor.fetchone()[0]
 
-            cursor.execute(SQL_GET_NETWORK_HABITAT_STATS, [area, network])
+            cursor.execute(SQL_GET_NETWORK_HABITAT_STATS, [total_area, network])
 
             while True:
                 try:
@@ -554,5 +555,9 @@ def habitat_statistics(request):
                 except ProgrammingError:
                     if not cursor.nextset():
                         break
+
+    unmapped_area = (total_area / 1000000) - float(sum([v['area'] for v in habitat_stats]))
+    unmapped_percentage = 100 * unmapped_area / (total_area / 1000000)
+    habitat_stats.append({'habitat': None, 'area': unmapped_area, 'percentage': unmapped_percentage})
 
     return Response(habitat_stats)
--- a/backend/habitat/viewsets.py
+++ b/backend/habitat/viewsets.py
@@ -93,6 +93,26 @@ SQL_GET_PARKS = "select name, network from SeamapAus_Parks_View;"
 SQL_GET_ZONES = "select name from SeamapAus_Zones_View;"
 SQL_GET_ZONES_IUCN = "select name from SeamapAus_ZonesIUCN_View;"
 
+SQL_GET_NETWORK_AREA = "select geom.STArea() from SeamapAus_Networks_View where name=%s;"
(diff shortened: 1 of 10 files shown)


backend/habitat/viewsets.py
```diff
@@ -490,3 +510,54 @@ def zones_iucn(request):
...
+    network   = request.query_params.get('network')
+    park      = request.query_params.get('park')
+    zone      = request.query_params.get('zone')
+    zone_iucn = request.query_params.get('zone-iucn')
+
+    habitat_stats = []
+    total_area = 0
+
+    with connections['transects'].cursor() as cursor:
+        if (park):
```
markhepburn, line 526: why the parentheses here?  (I don't think they do anything)

backend/habitat/viewsets.py
```diff
@@ -93,6 +93,26 @@
...
 SQL_GET_ZONES_IUCN = "select name from SeamapAus_ZonesIUCN_View;"
 
+SQL_GET_NETWORK_AREA = "select geom.STArea() from SeamapAus_Networks_View where name=%s;"
+SQL_GET_NETWORK_HABITAT_STATS = """
+select habitat, area / 1000000 as area, 100 * area / %s as percentage
+from SeamapAus_Habitat_By_Region
+where
+  region = %s and
+  boundary_layer_id = 260 and
+  habitat_layer_id = 95;
```
markhepburn, line 103: These are obviously the bits that jump out: it's _probably_ ok, at least for now, but let's keep it in mind while we look for a better way.

markhepburn: Actually, the habitat layer will be fixed I think (let's clarify this with Emma ASAP), so is that part of the view even needed?  (Similarly, for the boundary layer we can probably get that id from the user's selection)

(If this is WIP you may already have this in mind)

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -556,24 +556,49 @@
...
-(defn update-active-zone-iucn [db [_ zone-iucn]]
-  (-> db
-      (assoc-in [:map :active-zone-iucn] zone-iucn)
-      (assoc-in [:map :active-zone] nil)))
+(defn update-active-network [{:keys [db]} [_ network]]
+  (let [db (if-not (= network (get-in db [:map :active-network]))
+             (-> db
+                 (assoc-in [:map :active-park] nil)
+                 (assoc-in [:map :active-network] network))
+             db)]
```
markhepburn, line 564: This actually _is_ a good use-case for `cond->` I think!  (that general pattern of "if X do this transform, else just return the thing".  It's not so great for "if X do this transform, otherwise do this other transform").  And although it can be a bit tricky to follow at first, I'm sure I've seen `->` used as that transform; eg

```clojure
(cond-> db
  (not= network (get-in db [:map :active-network]))
  (->
    (assoc-in [:map :active-park] nil)
    (assoc-in [:map :active-network] network)))
```

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -556,24 +556,49 @@
...
+  (let [db (-> db
+               (assoc-in [:map :active-zone-iucn] zone-iucn)
+               (assoc-in [:map :active-zone] nil))]
+    {:db db
+     :dispatch [:map/get-habitat-statistics]}))
+
+(defn get-habitat-statistics [{:keys [db]}]
+  (let [habitat-statistics-url (get-in db [:config :habitat-statistics-url])
+        {:keys [active-network active-park active-zone active-zone-iucn]}                    (:map db)]
+   {:db db
```
markhepburn, line 592: It might not be obvious, but you don't actually _have_ to return a :db key here.  The :db is just another effect that says "swap out the app-db and replace it with this one", and if it isn't present the app-db will stay as it is)

backend/habitat/viewsets.py
JoshGx2000: Right you are! Removing them.

backend/habitat/viewsets.py
JoshGx2000: 100% WIP. This query borrows a lot from `SQL_GET_STATS` above it. I plan to have fixed views that would eliminate that part of the query, I just haven't gotten to it because I'm still planning out the zones stuff.

JoshGx2000: In fact, I might take a step back once I look at zones to see how I can include the necessary information in the four views I made yesterday.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: I actually only noticed this feature today. Anyway, I'll fix it.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Done, thanks!

PR #74, opened 2022-06-27
Feature: (ISA-163) Bathymetry UI & (ISA-161) Downloads query statistics

JoshGx2000:
[ISA-163](https://jira.its.utas.edu.au/browse/ISA-163)

Bathymetry statistics frontend added in state of knowledge drawer. Bathymetry statistics and habitat statistics can be collapsed/expanded to only view those that are relevant.

Extra query statistics added to both habitat and bathymetry statistics for the purposes of allowing downloads of each.


frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -866,53 +868,180 @@
...
+   [:a.data-coverage-report-link {:on-click open-data-coverage-report} "View data coverage report"]])
 
 (defn habitat-statistics
   []
-  (let [selected-tab (reagent/atom "breakdown")]
+  (let [selected-tab (reagent/atom "breakdown")
+        collapsed?   (reagent/atom false)]
     (fn []
-      (let [habitat-statistics @(re-frame/subscribe [:map/habitat-statistics])
+      (let [loading?           @(re-frame/subscribe [:map/habitat-statistics-loading?])
```
markhepburn, line 879: I'm not too fussed; this keeps the code simple, possibly at the cost of polluting the subscription namespace a bit.  An alternative might be to have a "union type" for the habitat stats; they're either :habitat-stats/loading _or_ the data (or I guess :habitat-stats/error, we probably ignore that for now but you get the gist).  As I said, not fussed here, just raising it as something to consider as this app gets larger.

markhepburn (approved):
I don't think this will merge automatically now, so if you wouldn't mind :)

PR #82, opened 2022-07-06
Bugfix: (ISA-187) Separate GetFeatureInfo request by info format

JoshGx2000:
[ISA-187](https://jira.its.utas.edu.au/browse/ISA-187)

GetFeatureInfo requests are separated by info format, allowing for mixed responses in a single popup.

Screen Shot 2022-07-06 at 3 16 48 pm
Screen Shot 2022-07-06 at 3 16 59 pm
Screen Shot 2022-07-06 at 3 17 13 pm


markhepburn (approved):
I think it should handle it, but has this been tested with a group of layers, in particular Emma's example of different formats on the same server?

PR #84, opened 2022-07-08
Feature: (ISA-161) Habitat observations queries

JoshGx2000:

~~Habitat observations in backend and used by frontend to update the app state. SQUIDLE observations query in the backend will need changes shortly when Emma comes through on some column updates.~~
Edit: SQUIDLE stuff has been updated and the habitat observations queries work as intended 🙂

### DIFF
--- a/backend/habitat/viewsets.py
+++ b/backend/habitat/viewsets.py
@@ -456,7 +456,7 @@ SELECT
   observation.METHOD AS method,
   observation.images,
   observation.total_annotations,
-  observation.public_annotations,
+  observation.public_annotations
 FROM (
   SELECT DISTINCT observation
   FROM BOUNDARY_AMP_HABITAT_OBS_SQUIDLE
@@ -480,7 +480,8 @@ SELECT
   observation.DATE AS date,
   observation.METHOD AS method,
   observation.images,
-  observation.annotations
+  observation.total_annotations,
+  observation.public_annotations
 FROM (
   SELECT DISTINCT observation
   FROM BOUNDARY_IMCRA_HABITAT_OBS_SQUIDLE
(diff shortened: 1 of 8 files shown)


backend/habitat/viewsets.py
```diff
@@ -870,3 +1083,82 @@ def bathymetry_statistics(request):
             return Response([])
         else:
             return Response(bathymetry_stats)
+
+@action(detail=False)
+@api_view()
+@renderer_classes((JSONRenderer, ShapefileRenderer))
```
markhepburn, line 1089: It's not a bad idea, but is there a concrete reason for ShapefileRenderer to be included here?  (more specifically, have you tested it?)

backend/habitat/viewsets.py
```diff
@@ -870,3 +1083,82 @@ def bathymetry_statistics(request):
...
+            raise Exception('Unhandled boundary type!')
+
+        try:
+            columns = [col[0] for col in cursor.description]
+            namedrow = namedtuple('Result', columns)
+            results = [namedrow(*row) for row in cursor.fetchall()]
+
+            global_archive = [row._asdict() for row in results]
+        except:
+            pass
```
markhepburn, line 1124: What's the thinking behind this?  Do we expect benign exceptions that we ignore, or is this just convenient?

(I'm not against it per-se, just wondering if any thought has gone into it, or it was just convenient at the time)

markhepburn (commented):
Looks good in general, just a couple of comments I left about general practice.

backend/habitat/viewsets.py
JoshGx2000: No reason, just a copy+paste error

backend/habitat/viewsets.py
JoshGx2000: Yeah, it's probably a good idea to not allow any benign exceptions. I wasn't really thinking. We'll just have it fail on any exception 👍

PR #86, opened 2022-07-11
Feature: (ISA-161 & ISA-164) Habitat observations summary UI

JoshGx2000:
[ISA-161](https://jira.its.utas.edu.au/browse/ISA-161) & [ISA-164](https://jira.its.utas.edu.au/browse/ISA-164)

Querying summary stats of habitat observations in backend and displaying them in the habitat observations UI.

Screen Shot 2022-07-11 at 12 49 45 pm
Screen Shot 2022-07-11 at 12 50 41 pm
Screen Shot 2022-07-11 at 12 50 37 pm

### DIFF
--- a/backend/habitat/viewsets.py
+++ b/backend/habitat/viewsets.py
@@ -306,11 +306,23 @@ GROUP BY bathymetry_resolution, bathymetry_rank;
 """
 
 SQL_GET_AMP_HABITAT_OBS_GLOBALARCHIVE = """
+SET NOCOUNT ON;
+
 DECLARE @netname  NVARCHAR(254) = %s;
 DECLARE @resname  NVARCHAR(254) = %s;
 DECLARE @zonename NVARCHAR(254) = %s;
 DECLARE @zoneiucn NVARCHAR(5)   = %s;
 
+DECLARE @observations TABLE
+(
+  campaign_name NVARCHAR(MAX) NOT NULL, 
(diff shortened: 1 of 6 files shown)


backend/habitat/viewsets.py
```diff
@@ -1099,57 +1286,57 @@ def habitat_observations(request):
...
+    sediment = None
+    squidle = None
 
     with connections['transects'].cursor() as cursor:
         try:
-            # Global Archives observations
+            # Global Archives stats
             if boundary_type == 'amp':
-                cursor.execute(SQL_GET_AMP_HABITAT_OBS_GLOBALARCHIVE, [network, park, zone, zone_iucn])
+                cursor.execute(SQL_GET_AMP_HABITAT_OBS_GLOBALARCHIVE + SQL_GET_GLOBALARCHIVE_STATS, [network, park, zone, zone_iucn])
```
markhepburn, line 1297: It feels vaguely dirty just munging sql strings like this, particularly when there's a shared data structure involved, and yet... it also seems like a nice efficient way to go!

markhepburn (approved):
I just thought to check -- that views file is 1500 lines now, might be worth pulling the state-of-knowledge stuff out into a separate module!

backend/habitat/viewsets.py
JoshGx2000: Yeah, it'd be great if I could somehow signify some query constants as fragments but 🤷

PR #90, opened 2022-07-15
Feature: (ISA-55) Layer preview image generation

JoshGx2000:
[ISA-55](https://jira.its.utas.edu.au/browse/ISA-55)

Real preview images of layers are now generated and shown when hovered over layers:
Screen Shot 2022-07-15 at 10 38 57 am


backend/catalogue/management/commands/generate_layer_previews.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+    def handle(self, *args, **options):
+        layer_id = int(options['layer_id']
+                       ) if options['layer_id'] != None else None
+
+        if layer_id is not None:
+            layer = Layer.objects.get(id=layer_id)
+
+            bytes_io = BytesIO()
```
markhepburn, line 91: Minor point of style -- have you see the python "with" statement?  (in this scenario it would basically be a try-finally; automatically closing the resource)

backend/webapp/settings.py
```diff
@@ -138,3 +138,7 @@
 # Use to override the OPTIONS.driver value in the database settings
 # (eg, if django is using a different driver to ogr2ogr)
 OGR2OGR_DRIVER = None
+
+
+MEDIA_ROOT = 'media/'
+MEDIA_URL = 'media/'
```
markhepburn, line 144: These bits make me nervous -- I _think_ they should already be set or have default values, but also it looks like your'e treating them as the same thing.  MEDIA_ROOT is the disk/storage location, while MEDIA_URL is the server endpoint. (Having typed all that you're probably doing the right thing, it just looks odd and worth commenting on)

frontend/shadow-cljs.edn
```diff
@@ -29,19 +29,21 @@
                                    imas-seamap.specs.preload]}
 
           :min {:target            :browser
-                :closure-defines   {goog.DEBUG                  false
-                                    imas-seamap.db/api-url-base "https://data.imas.utas.edu.au/seamap/api/"
-                                    imas-seamap.db/img-url-base "/app/img/"}
+                :closure-defines   {goog.DEBUG                    false
+                                    imas-seamap.db/api-url-base   "https://data.imas.utas.edu.au/seamap/api/"
+                                    imas-seamap.db/media-url-base "https://data.imas.utas.edu.au/seamap/media/"
```
markhepburn, line 34: Aah, yeah I'm not sure about this.  It would be nice if we could avoid that here; worth a thought (I guess the frontend needs the url base... hmm)

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -800,17 +800,11 @@
...
-    [:div.layer-preview
-     [:img
-      {:src (case (mod (:id preview-layer) 5) ; Selects one of five placeholder images based on layer id - TODO: Replace with actual layer preview image per layer and remove placeholder images from project
-              0 "img/LayerPreview1.png"
-              1 "img/LayerPreview2.png"
-              2 "img/LayerPreview3.png"
-              3 "img/LayerPreview4.png"
-              4 "img/LayerPreview5.png")}]]))
+(defn layer-preview []
+  (let [preview-layer-url @(re-frame/subscribe [:ui/preview-layer-url])]
```
markhepburn, line 804: Random aside; there's a `when-let`

frontend/shadow-cljs.edn
markhepburn: There's a few hacky values like that now, I wonder if it's worth injecting in an initial bootstrap payload... not the most important thing, it's just looking ugly.

markhepburn (approved):

backend/webapp/settings.py
markhepburn: Actually on deploying I found an issue -- it assumes that basemap is in the media directory, but it isn't (at least, not the production configured media root)

PR #93, opened 2022-07-25
Feature: (ISA-192) Generate survey graphs

JoshGx2000:
Retrieves survey data from the database, and then generates graphs showing the number of surveys conducted over time.
Generates graphs showing data for all years, and only data post-2000.
South-west (2000 onwards)

Generated graphs stored in media directory.

### DIFF
--- a/backend/catalogue/management/commands/generate_survey_graphs.py
+++ b/backend/catalogue/management/commands/generate_survey_graphs.py
@@ -1,0 +1,172 @@
+from django.core.management.base import BaseCommand
+from django.core.files.storage import default_storage
+from django.core.files import File
+from django.db import connections
+from collections import namedtuple
+import matplotlib.pyplot as plt
+from matplotlib.ticker import MaxNLocator
+from io import BytesIO
+
+# groups the list of data by network
+def group_networks(data):
+    networks = set(map(lambda v: v.get('NETWORK'), data))
+    return [{'network': network, 'value': [v for v in data if v.get('NETWORK') == network]} for network in networks]
+
+
+# groups the list of data by park
+def group_parks(data):
+    parks = set(map(lambda v: v.get('PARK'), data))
+    return [{'park': park, 'value': [v for v in data if v.get('PARK') == park]} for park in parks]
+
+
+# summarises the list of data by each year
+def summarise_by_year(data):
+    years = set(map(lambda v: v.get('YEAR'), data))
+    return [{
+        'year': year,
+        'imagery_count': sum(v.get('imagery_count') for v in data if v.get('YEAR') == year),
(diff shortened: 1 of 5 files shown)


backend/catalogue/management/commands/generate_survey_graphs.py
```diff
@@ -0,0 +1,172 @@
...
+    filename = title
+
+    if exclude_old_data:
+        min_year = max(min_year, 2000)
+        title = title + ' (2000 onwards)'
+
+    years = [v for v in range(min_year, max_year + 1)]
+
+    imagery = [
+        next((v.get('imagery_count') for v in data if v.get('year') == year), 0)
```
markhepburn, line 46: I'm trying to nut this out: do you have duplicate years?  (It just seems a bit weird to be creating an iterator like that and calling next() on it.  I can see that it works.. just wondering if there's a better or clearer way).

Can you sort `data` by `year`, or even just turn data into a dict indexed by year?  (Now I'm needlessly looking for efficiency and away from clarity, but I bet there's an itertools function that would turn `[dict(video,sediment,etc)]` into `[[video],[sediment],..]`!)

markhepburn: (It's basically a pivot, isn't it!  Turn a list of dicts with known keys and singular vals, into a dict of those keys and lists of vals.  Handling missing years might get tricky.  Anyway I'm rambling now)

backend/catalogue/management/commands/generate_survey_graphs.py
```diff
@@ -0,0 +1,172 @@
+from django.core.management.base import BaseCommand
+from django.core.files.storage import default_storage
+from django.core.files import File
+from django.db import connections
+from collections import namedtuple
+import matplotlib.pyplot as plt
+from matplotlib.ticker import MaxNLocator
+from io import BytesIO
```
markhepburn, line 8: Probably a useful habit to get into by including logging (a chat I've been having with Oli lately)

markhepburn (commented):
Looks good, but might be worth talking through a couple of the points in there.

backend/catalogue/management/commands/generate_survey_graphs.py
JoshGx2000: As discussed in Google Chat, this was just me forgetting stuff and being dumb. I've fixed this now so there's _way_ less redundant code, and the readability has greatly improved.

backend/catalogue/management/commands/generate_survey_graphs.py
JoshGx2000: Done!

PR #95, opened 2022-07-27
Feature: (ISA-55) Layer preview image improvements

JoshGx2000:
A few fixes to the image generation code, and code for a fallback on the frontend if the image retrieval fails:
Screen Shot 2022-07-27 at 1 27 44 pm


backend/catalogue/management/commands/generate_layer_previews.py
```diff
@@ -78,42 +92,58 @@ def retrieve_image(layer):
...
+            '--only_generate_missing',
+            help='If true, skips generating images for layers where images already exist'
         )
 
     def handle(self, *args, **options):
-        layer_id = int(options['layer_id']
-                       ) if options['layer_id'] != None else None
+        layer_id = int(options['layer_id']) if options['layer_id'] != None else None
+        only_generate_missing = options['only_generate_missing'].lower() in ['t', 'true'] if options['only_generate_missing'] != None else False
+
```
markhepburn, line 106: Trivial, but there's a tonne of whitespace here!

backend/catalogue/management/commands/generate_layer_previews.py
```diff
@@ -78,42 +92,58 @@ def retrieve_image(layer):
...
+
 
         if layer_id is not None:
             layer = Layer.objects.get(id=layer_id)
 
-            bytes_io = BytesIO()
+            filepath = f'layer_previews/{layer.id}.png'
+            if not only_generate_missing or not default_storage.exists(filepath):
+                bytes_io = BytesIO()
```
markhepburn, line 114: I'd recommend using a with-block for bytesio (just avoids traps with cleanup, etc).  Bit of a dry reference, but these: https://docs.python.org/3/reference/compound_stmts.html#with

markhepburn (approved):
Looks good -- I'd suggest getting into the habit of using with-blocks in python though (think of them as try-finally; typically just closes resources for you, but safely and idiomatically)

backend/catalogue/management/commands/generate_layer_previews.py
JoshGx2000: Ah, very useful to know! I guess I should have been more curious about that syntax when it was being used for the database connection. I've used it as suggested.

backend/catalogue/management/commands/generate_layer_previews.py
JoshGx2000: Fixed!

PR #97, opened 2022-07-29
Feature: (ISA-205) State of knowledge boundary CQL filters

JoshGx2000:
[ISA-205](https://jira.its.utas.edu.au/browse/ISA-205)

Using CQL filters on the active boundary layer to show which area is currently being filtered to in the state of knowledge.
Appropriate layer that corresponds with filters is now shown (i.e. showing zones layer when zone filter is selected, park filter when park filter is select, etc.)

Screen Shot 2022-07-29 at 1 56 54 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/views.cljs
+++ b/frontend/src/cljs/imas_seamap/map/views.cljs
@@ -164,6 +164,7 @@
         download-info                                 @(re-frame/subscribe [:download/info])
         layer-priorities                              @(re-frame/subscribe [:map.layers/priorities])
         ;layer-params                                  @(re-frame/subscribe [:map.layers/params])
+        boundary-filter                               @(re-frame/subscribe [:sok/boundary-layer-filter])
         logic-type                                    @(re-frame/subscribe [:map.layers/logic])
         loading?                                      @(re-frame/subscribe [:app/loading?])]
     (into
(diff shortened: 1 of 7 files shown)


frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -217,7 +218,7 @@
       ;; then orders in the map by update not by list):
        (map-indexed
         (fn [i {:keys [server_url layer_name style id] :as layer}]
-          ^{:key (str id)}
+          ^{:key (str id (boundary-filter layer))}
```
markhepburn, line 221: Good job on finding this (not sure how long it would have taken me!), but I'm curious about it now.  It's effectively saying that layers are considered immutable by react-leaflet I think (which might be a bug?); they won't re-render just because those props have changed, you need to force it by giving it a new id.

(good lord RL has changed a bit!  It's all react hooks and contexts and providers now; I had a patch or two incorporated in the early versions, and it was fairly easy to understand how it translated react prop changes to the underlying leaflet element.  Now... I have no idea)

markhepburn (approved):

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Thanks 😅

PR #98, opened 2022-08-01
(ISA-215) Django logging config

JoshGx2000:
[ISA-215](https://jira.its.utas.edu.au/browse/ISA-215)

Added logging config to settings.py.
@markhepburn mentioned something about journalctl in the ticket; can I get more info on this?


markhepburn:
More accurately "journald", which manages the log files etc, and journalctl is the command to manage and view logs.

So: systemd (and frontend systemctl) is the init system in most linux systems these days; it's the first process started, and controls all the services that run (including for our purposes, the seamap API and geoserver et al).  That means defining when they start and if they have any dependencies, the user they run as and the working directory, and what to do with stdout/stderr.  Each service has a "unit file", with extension ".service", defining this.  Most are system unit definitions, but you can also have user ones, eg have a look in ~/.config/systemd/user/ in imasdata-dev1.  User services are managed with the --user argument, eg "systemctl --user restart tomcat_geoserver".  Basic overview eg https://www.redhat.com/sysadmin/getting-started-systemctl

Journald is part of systemd, and manages logs.  To view them it's a similar situation; eg "journalctl --user -u seamap.service" (I need the .service extension because there's a seamap.socket file as well).  That only works if you're using journald though, which basically means logging to stdout and stderr.  I've just been going through that extercise myself, because systemd actually starts gunicorn which runs django, so there's a layer in the middle.  Coupling that with logs that get excluded if DEBUG is false, etc, and it's tricky.

So what you have is a great start; how have you tested it?  (And what's the final "effective" configuration; eg would any errors get logged?)

JoshGx2000:
Currently all the logging is done to the console and nothing more. I've tested it just by running one of the management commands (`generate_layer_previews`). A lot of what you've said above is above my head, so I'll need to take some time to research the stuff you've mentioned.

markhepburn:
My main question (the rest was just background info since you asked); what's the effective configuration for django?  My understanding is those get merged with the existing defaults, providing override_existing_loggers:False, or whatever the setting is.

PR #100, opened 2022-08-03
Feature: (ISA-218) Zoom selected bounds

JoshGx2000:
[ISA-218](https://jira.its.utas.edu.au/browse/ISA-218)

Automatically zooms to the bounds of the selected filters for a state of knowledge boundary. It's a little slow because the WFS GetFeature request is asynchronous, so sometimes it feels weird – I'll let others decide if it feels too weird for users though.

Screen Shot 2022-08-03 at 4 01 36 pm
Screen Shot 2022-08-03 at 4 01 42 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/state_of_knowledge/utils.cljs
+++ b/frontend/src/cljs/imas_seamap/state_of_knowledge/utils.cljs
@@ -26,3 +26,21 @@
      :realm                (:name active-realm)
      :province             (:name active-province)
      :ecoregion            (:name active-ecoregion)}))
+
+(defn cql-filter [boundaries]
+  (let [{:keys
+         [network park zone zone-iucn
+          provincial-bioregion mesoscale-bioregion
+          realm province ecoregion]} (boundary-filter-names boundaries)
+        filters (remove
+                 nil?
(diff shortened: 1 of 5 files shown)


markhepburn:
Yeah, I wondered about any noticeable lag (which is probably going to bew worse in the real world); I wonder if there's some bit of feedback we could incorporate, subtle but so that people can learn that something might still be happening..

frontend/src/cljs/imas_seamap/state_of_knowledge/utils.cljs
```diff
@@ -26,3 +26,21 @@
      :realm                (:name active-realm)
      :province             (:name active-province)
      :ecoregion            (:name active-ecoregion)}))
+
+(defn cql-filter [boundaries]
+  (let [{:keys
+         [network park zone zone-iucn
+          provincial-bioregion mesoscale-bioregion
+          realm province ecoregion]} (boundary-filter-names boundaries)
+        filters (remove
```
markhepburn, line 35: This makes me want to code-golf it!  It's fine, but the use of both `when` and `remove-nil` makes it look redundant.  I can't think of a clearer way though..

markhepburn (approved):

PR #102, opened 2022-08-04
Feature: (ISA-222) Replace magic mode

JoshGx2000:
[ISA-222](https://jira.its.utas.edu.au/browse/ISA-222)

"Magic mode" has been replaced with an option that only shows layers within the current viewport.
Screen Shot 2022-08-04 at 4 29 01 pm
Screen Shot 2022-08-04 at 4 29 05 pm
Screen Shot 2022-08-04 at 4 29 18 pm

Need to figure out how we're going have the national layer displayed on load – hardcode it as suggested by Emma?

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/utils.cljs
+++ b/frontend/src/cljs/imas_seamap/map/utils.cljs
@@ -232,12 +232,29 @@
   [coll]
   (sort-by (juxt #(or (:sort_key %) "zzzzzzzzzz") :id) coll))
 
-(defn layer-visible? [{:keys [west south east north] :as _bounds}
-                      {:keys [bounding_box]          :as _layer}]
-  (not (or (> (:south bounding_box) north)
-           (< (:north bounding_box) south)
-           (> (:west  bounding_box) east)
-           (< (:east  bounding_box) west))))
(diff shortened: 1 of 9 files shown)


frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -255,3 +231,30 @@
   "Sorts a collection by its sort-key first and its id second."
   [coll]
   (sort-by (juxt #(or (:sort_key %) "zzzzzzzzzz") :id) coll))
+
+(defn normal-latitude
+  "Latitude can get pretty wacky if one loops around the entire globe a few times.
+   This puts the latitude within the normal latitude range."
+  [lat]
+  (-> lat
+   (+ 180)
```
markhepburn, line 240: Trivial, but this indentation feels off

markhepburn (approved):

frontend/src/cljs/imas_seamap/map/utils.cljs
JoshGx2000: Whoops, yep. Trivial though. I have an idea of how it happened.

PR #103, opened 2022-08-05
Feature: (ISA-221) Include server-set CSS in GetFeatureInfo popup

JoshGx2000:
[ISA-221](https://jira.its.utas.edu.au/browse/ISA-221)

Thanks to the encapsulation provided by [Shadow-DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM), we can now include the style from GetFeatureInfo responses in addition to their body content. This has fixed the GFI display for a few layers.

Before:
Screen Shot 2022-08-05 at 5 22 23 pm
After:
Screen Shot 2022-08-05 at 5 22 37 pm

### DIFF
--- a/backend/catalogue/management/commands/generate_layer_previews.py
+++ b/backend/catalogue/management/commands/generate_layer_previews.py
@@ -10,8 +10,9 @@ import logging
 
 from catalogue.models import Layer
 
+Image.MAX_IMAGE_PIXELS = None
 
-basemap = Image.open(default_storage.open('basemap.png'))
+basemap = Image.open(default_storage.open('land_shallow_topo_21600.tif'))
 
 
 def basemap_latitude_to_pixel(latitude):
--- a/frontend/src/cljs/imas_seamap/map/utils.cljs
+++ b/frontend/src/cljs/imas_seamap/map/utils.cljs
@@ -222,19 +222,22 @@
         str)))
 
 (def feature-info-none
-  {:body
+  {:style nil
+   :body
    (str
     "<div>"
-    "<h4>No info available</h4>"
-    "Layer summary not configured"
+    "    <h4>No info available</h4>"
+    "    Layer summary not configured"
     "</div>")})
 
 (defn feature-info-html
   [response]
   (let [parsed (.parseFromString (js/DOMParser.) response "text/html")
-        body  (first (array-seq (.querySelectorAll parsed "body")))]
+        body  (first (array-seq (.querySelectorAll parsed "body")))
+        style  (first (array-seq (.querySelectorAll parsed "style")))] ; only grabs the first style element
     (when (.-firstElementChild body)
-      {:body (.-innerHTML body)})))
+      {:style (when style (.-innerHTML style))
+       :body (.-innerHTML body)})))
 
 (defn feature-info-json
   [response]
--- a/frontend/src/cljs/imas_seamap/map/utils.cljs
+++ b/frontend/src/cljs/imas_seamap/map/utils.cljs
@@ -244,11 +244,31 @@
         property-to-row (fn [{:keys [label value]}] (str "<tr><td>" label "</td><td>" value "</td></tr>"))
         property-rows (str/join "" (map (fn [property] (property-to-row property)) properties))]
     (when (or id (not-empty properties))
-      {:body (str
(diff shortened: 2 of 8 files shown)


frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -198,19 +198,22 @@
...
 
 (defn feature-info-html
   [response]
   (let [parsed (.parseFromString (js/DOMParser.) response "text/html")
-        body  (first (array-seq (.querySelectorAll parsed "body")))]
-    (if (.-firstElementChild body)
-      {:info (.-innerHTML body)}
-      {:status :feature-info/empty})))
+        body  (first (array-seq (.querySelectorAll parsed "body")))
+        style  (first (array-seq (.querySelectorAll parsed "style")))] ; only grabs the first style element
```
markhepburn, line 213: Yeah I was going to look for this -- I suppose in the general instance it's not ideal, but for basic GFI responses that probably just have a single style block it's probably fine.

frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -219,13 +222,32 @@
...
-              "<table>" property-rows "</table>"
-              "</div>")}
-      {:status :feature-info/empty})))
+    (when (or id (not-empty properties))
+      {:style
+       (str
+        ".feature-info-json {"
+        "    max-height: 257px;"
+        "    overflow-y: scroll;"
+        "}"
```
markhepburn, line 231: Don't know why I didn't do it myself, but worth noting that clojure has multi-line strings out of the box.

frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -202,7 +202,13 @@
...
                                    :icon  "error"}]
     ;; Default; we have actual content:
-    [:div {:dangerouslySetInnerHTML {:__html info-body}}]))
+    [:div
+     {:ref
+      (fn [element]
+        (when element
+          (set! (.-innerHTML element) nil)
+          (doseq [{:keys [_body _style] :as response} responses]
+            (.appendChild element (create-shadow-dom-element response)))))}]))
```
markhepburn, line 211: What's the deal with refs, out of curiosity -- the only thing I'd want to ensure here is that the function isn't going to be called repeatedly (which I assume it doesn't, I've just never looked into it)

backend/catalogue/management/commands/generate_layer_previews.py
```diff
@@ -10,8 +10,9 @@
 
 from catalogue.models import Layer
 
+Image.MAX_IMAGE_PIXELS = None
 
-basemap = Image.open(default_storage.open('basemap.png'))
+basemap = Image.open(default_storage.open('land_shallow_topo_21600.tif'))
```
markhepburn, line 15: Just an FYI, looks like this crept in.

markhepburn (approved):

frontend/src/cljs/imas_seamap/map/utils.cljs
JoshGx2000: Yeah, I noticed that, but it has mucked up indentation, at least in my editor. I'll use them in future, knowing they are preferred.

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Once per state change, at least that's what my testing shows.

backend/catalogue/management/commands/generate_layer_previews.py
JoshGx2000: Hmm, yeah. I thought I had directly committed this on the develop branch as part of a separate ticket. Not sure if I accidentally committed on this branch, or whether things got a bit confused when I merged develop into this branch.

frontend/src/cljs/imas_seamap/map/utils.cljs
JoshGx2000: Yep. If it comes down to it I can always right extra code to merge the contents of all style elements (but this was the simplest solution).

PR #105, opened 2022-08-09
Feature: (ISA-222, ISA-223, & ISA-225) Keyed-layers

JoshGx2000:
[ISA-222](https://jira.its.utas.edu.au/browse/ISA-222), [ISA-223](https://jira.its.utas.edu.au/browse/ISA-223), [ISA-225](https://jira.its.utas.edu.au/browse/ISA-225)

It's become increasingly clear that there's a need to be able to have "hardcoded" layers in the app for certain situations (e.g. default layer(s) on startup). This is a problem if IMAS needs to swap out what hardcoded layers are used in the app if they update the layers in the future.

The solution to this is having a fixed-length table with discrete keywords that alias layers that can then be used in the app instead of directly referring to a layer's ID. What this means is that if IMAS needs to switch out a hardcoded layer in the app, all they need to do is update the keyed-layers table.

### DIFF
--- a/database/Procedures/UpdateBathymetry.sql
+++ b/database/Procedures/UpdateBathymetry.sql
@@ -26,6 +26,12 @@ BEGIN
     [Group]      INT         NOT NULL
   );
 
+  DROP TABLE IF EXISTS ##T4;
+  CREATE TABLE ##T4 (
+    [geom]       GEOMETRY    NOT NULL,
+    [Group]      INT         NOT NULL
+  );
+
   DROP TABLE IF EXISTS ##Groups;
   CREATE TABLE ##Groups (
     [geom]       GEOMETRY    NOT NULL
@@ -38,7 +44,7 @@ BEGIN
       ROW_NUMBER() OVER(
         ORDER BY [RESOLUTION], [RANK], [ID]
       ) - 1
-    ) / 15
+    ) / 10
   FROM [dbo].[VW_BATHYMETRY]
   WHERE [RESOLUTION] = @resolution;
   
@@ -60,13 +66,24 @@ BEGIN
       ROW_NUMBER() OVER(
         ORDER BY [Group]
       ) - 1
-    ) / 5
+    ) / 3
   FROM ##T2
+  GROUP BY [Group];
+
+  INSERT INTO ##T4
+  SELECT
+    GEOMETRY::UnionAggregate([geom]),
+    (
+      ROW_NUMBER() OVER(
+        ORDER BY [Group]
+      ) - 1
+    ) / 3
+  FROM ##T3
   GROUP BY [Group];
 
   INSERT INTO ##Groups
   SELECT GEOMETRY::UnionAggregate([geom])
-  FROM ##T3
+  FROM ##T4
   GROUP BY [Group];
 
   DECLARE @geom GEOMETRY = (
--- a/frontend/src/cljs/imas_seamap/events.cljs
+++ b/frontend/src/cljs/imas_seamap/events.cljs
@@ -109,7 +109,7 @@
                   base-layer-groups
                   grouped-base-layers groups
                   priorities organisations
-                  categories]
+                  categories keyed-layers]
                  :as _map-state} :map
                 {{{:keys [networks parks zones zones-iucn]} :amp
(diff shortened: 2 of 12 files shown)


JoshGx2000:
Converted to a draft because some of the features regarding keyed layers need to be sorted out.

JoshGx2000:
Okay, now everything should account for non-unique keys

frontend/src/cljs/imas_seamap/events.cljs
```diff
@@ -140,7 +141,7 @@
                (merge {:habitat-colours habitat-colours
                        :habitat-titles  habitat-titles
                        :sorting         sorting}))]
-    (assoc-in db [:map :active-layers] (vec (applicable-layers db :category :habitat)))))
+    (assoc-in db [:map :active-layers] (or (get-in db [:map :keyed-layers :startup]) []))))
```
markhepburn, line 144: `get-in` has a default-value option, no need for `or`

database/Procedures/UpdateBathymetry.sql
```diff
@@ -26,6 +26,12 @@ BEGIN
     [Group]      INT         NOT NULL
   );
 
+  DROP TABLE IF EXISTS ##T4;
+  CREATE TABLE ##T4 (
+    [geom]       GEOMETRY    NOT NULL,
+    [Group]      INT         NOT NULL
+  );
+
```
markhepburn, line 34: Should this file be included?

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -277,12 +277,25 @@
   (let [db (assoc-in db [:map :base-layers] layers)]
     (update-grouped-base-layers db)))
 
+(defn- keyed-layers-join
+  "Using layers and keyed-layers, replaces layer IDs in keyed-layers with layer
+   objects. We only update keyed-layers if the db contains layers."
```
markhepburn, line 282: That comment made me think (because yes, race conditions have been a factor before).  There's no real reason to expand these into layers up-front, it just means the sub becomes "return this app-db path", but you could defer that to sub-time if you want.

markhepburn:
I just had to triple-check that it's not covid-brain, but there's a lot of overlap between #105 and #106! (eg the django side of things)

frontend/src/cljs/imas_seamap/events.cljs
JoshGx2000: Ah yep, thanks!

database/Procedures/UpdateBathymetry.sql
JoshGx2000: I am confused. This change was made on the develop branch, which was then merged into this branch. I have no idea why it's showing up in the file changes.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: > I just had to triple-check that it's not covid-brain, but there's a lot of overlap between #105 and #106! (eg the django side of things)

Yeah, I added a comment on #106 where I said that this PR was a prerequisite because I needed some of the code functionality. I'm now noting that it's causing some issues when you're reviewing the file diffs of the pull requests because it's saying the code changes are in both. In future I'll just branch off these feature branches if I need features that haven't yet been merged into develop (unfortunately I already made PR #108, so you're going to see weird code diffs in that that look like #107).

JoshGx2000:
> I just had to triple-check that it's not covid-brain, but there's a lot of overlap between #105 and #106! (eg the django side of things)

Yeah, I added a comment on #106 where I said that this PR was a prerequisite because I needed some of the code functionality. I'm now noting that it's causing some issues when you're reviewing the file diffs of the pull requests because it's saying the code changes are in both. In future I'll just branch off these feature branches if I need features that haven't yet been merged into develop (unfortunately I already made PR #108, so you're going to see weird code diffs in that that look like #107).

database/Procedures/UpdateBathymetry.sql
JoshGx2000: **Addendum:** See my comment at the bottom of this PR.

PR #107, opened 2022-08-11
Feature: (ISA-233) Popup panning

JoshGx2000:
[ISA-233](https://jira.its.utas.edu.au/browse/ISA-233)

Pans to a popup once loaded if that popup will be out of bounds. Doesn't pan if the entire popup is within the bounds of the map view. Only pans the amount needed to show the entire popup in the view.

Screen Shot 2022-08-11 at 12 08 59 pm
Screen Shot 2022-08-11 at 12 09 07 pm
Screen Shot 2022-08-11 at 12 09 16 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -114,7 +114,6 @@
                           :feature
                           {:status   :feature-info/waiting
                            :location point}))
-        db             (assoc-in db [:map :center] point)
         db             (assoc-in db [:feature :status] :feature-info/waiting)]
     (merge
      {:db db}
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -501,3 +501,27 @@
   (let [db (update-in db [:map :viewport-only?] not)]
     {:db       db
      :put-hash (encode-state db)}))
+
+(defn pan-to-popup
+  "Based on a known location and size of the popup, we can find out if it will be
+   outside of the bounds of the map. Knowing this we can pan the map to fix this."
+  [{{{{popup-x :x popup-y :y popup-lat :lat popup-lng :lng} :location} :feature
+    {{map-width :x map-height :y} :size [map-lat map-lng] :center}
+    :map :as db} :db}
+   [_ {popup-width :x popup-height :y :as _popup-dimensions}]]
+  (let [map-x           (/ map-width 2)
+        map-y           (/ map-height 2)
+        overflow-top    (- popup-height popup-y) ; positive values mean we're over the bounds by that many pixels (which tells us how many pixels we'd need to move the map in the opposite direction to have everything in-bounds)
+        overflow-bottom (- popup-y map-height)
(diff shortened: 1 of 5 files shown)


frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -420,19 +419,22 @@
...
      :dispatch [:ui/hide-loading]}))
 
-(defn map-view-updated [{:keys [db]} [_ {:keys [zoom center bounds]}]]
+(defn map-view-updated [{:keys [db]} [_ {:keys [zoom size center bounds]}]]
   ;; Race-condition warning: this is called when the map changes to
   ;; keep state synchronised, but the map can start generating events
   ;; before the rest of the app is ready... avoid this by flagging
   ;; initialised state:
-  (when (:initialised db)
+  ;; Exceptions apply for size because we're only reading that from the map
```
markhepburn, line 427: Does it matter that we set the size before it's initialised?

frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -192,23 +201,27 @@
                                   (merge
                                    {:title       "No Results"
                                     :description "Try clicking elsewhere or adding another layer"
-                                    :icon        "warning-sign"}
+                                    :icon        "warning-sign"
+                                    :ref         #(when % (re-frame/dispatch [:map/pan-to-popup {:x 305 :y 210}]))}
```
markhepburn, line 205: So what are these; hard-coded popup dimensions?

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Yes, it's necessary. If don't update it when the map is first created, then the next update won't happen until the user moves the map or resizes the window.
The map size is important because we need the size of the map to [calculate popup overflow](https://github.com/IMASau/Seamap/blob/f74698e30ba4281421e3f03a7ee4af0ba946c690/frontend/src/cljs/imas_seamap/map/events.cljs#L515-L517) and [how much to pan the map for each pixel of overflow](https://github.com/IMASau/Seamap/blob/f74698e30ba4281421e3f03a7ee4af0ba946c690/frontend/src/cljs/imas_seamap/map/events.cljs#L518-L519).
If the size of the map hasn't yet been set then we get some odd behaviour, like panning an infinite distance.

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Yeah. Unlike the retrieved results which are HTML elements, the feature-status popups are React elements which I can't get the width/height of. It just so happens however that their size is consistent, so there's no issue with just using a hardcoded size.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Added comment to code for clarity.

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Added comment to code for clarity.

PR #116, opened 2022-08-30
Feature: (ISA-256) Vector legends

JoshGx2000:
[ISA-256](https://jira.its.utas.edu.au/browse/ISA-256)

Now using vector legends instead of images.

Layers with a custom legend image url (e.g. MODELLED - Aus Bathymetry Grid 2019 - 250m [GA]) will still use the image from the url provided – open to opinions on this.

Before:
Screen Shot 2022-08-30 at 1 39 41 pm
After:
Screen Shot 2022-08-30 at 1 39 47 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/core.cljs
+++ b/frontend/src/cljs/imas_seamap/core.cljs
@@ -39,6 +39,7 @@
     :map.layers/lookup                    msubs/map-layer-lookup
     ;:map.layers/params                    msubs/map-layer-extra-params-fn
     :map.layer/info                       subs/map-layer-info
+    :map.layer/legend                     msubs/layer-legend
     :map.layer.selection/info             msubs/layer-selection-info
     :map.feature/info                     subs/feature-info
     ;:map/region-stats                     msubs/region-stats
--- a/frontend/src/cljs/imas_seamap/core.cljs
+++ b/frontend/src/cljs/imas_seamap/core.cljs
@@ -153,7 +153,7 @@
     :map.region-stats/select-habitat      mevents/region-stats-select-habitat
     :map/update-base-layers               mevents/update-base-layers
     :map/update-base-layer-groups         mevents/update-base-layer-groups
-    :map/update-layers                    mevents/update-layers
+    :map/update-layers                    [mevents/update-layers]
     :map/update-organisations             mevents/update-organisations
     :map/update-classifications           mevents/update-classifications
     :map/update-descriptors               mevents/update-descriptors
--- a/frontend/src/cljs/imas_seamap/core.cljs
+++ b/frontend/src/cljs/imas_seamap/core.cljs
@@ -141,6 +141,8 @@
     :map.layer/opacity-changed            [mevents/layer-set-opacity]
     :map.layers/filter                    [mevents/map-set-layer-filter]
     :map.layers/others-filter             mevents/map-set-others-layer-filter
+    :map.layer/get-legend                 [mevents/get-layer-legend]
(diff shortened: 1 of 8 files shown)


markhepburn:
(before starting, I think using the alternative url is the right call, that's literally what it's there for)

frontend/src/cljs/imas_seamap/core.cljs
```diff
@@ -141,6 +142,8 @@
     :map.layer/opacity-changed            [mevents/layer-set-opacity]
     :map.layers/filter                    [mevents/map-set-layer-filter]
     :map.layers/others-filter             mevents/map-set-others-layer-filter
+    :map.layer/get-legend                 [mevents/get-layer-legend]
+    :map.layer/got-legend                 mevents/got-layer-legend
```
markhepburn, line 146: Not suggesting we change at this point, but I wonder if this is the sort of thing that would be better off as a component?  You could use the async programming features, and not have to reframe namespace with all the callback events.  Possibly one to talk through with OG.

frontend/src/cljs/imas_seamap/views.cljs
```diff
@@ -100,21 +100,30 @@
...
+  (let [{:keys [has-info? loading? info]} @(re-frame/subscribe [:map.layer/legend layer])]
     [:div.legend-wrapper
-     [:img {:src legend-url}]]))
+     (cond
+       legend_url [:img {:src legend_url}] ; if we have a custom legend url, use that to display an image
+       has-info?  [vector-legend info]     ; else if we have the info for the layer then display it
+       loading?   [b/spinner]              ; else if we're loading show that
+       :else      (do                      ; else dispatch a request for the legend info
+                    (re-frame/dispatch [:map.layer/get-legend layer])
+                    [b/spinner]))]))
```
markhepburn, line 126: I'm a bit fried and not instantly seeing the alternative, but this side-effect is making me feel icky.  (it's clever and I can see that it would work... I just want a cleaner way :))

markhepburn: @olivergeorge mind weighing in here?  The problem is we want a component that displays a legend, but where previously the legend was just an img url, if we want vector legend support we have to make an ajax request.  Where do you think the best place to trigger that event is?  (I was musing in another part of the code that it might be better as a js component, rather than a bunch of reframe callback events.  That's still fairly side-effecty, so perhaps a wrapper that retrieves the contents, then a dumb rendering component?)

markhepburn (changes requested):
Good start, although I'm nudging Oli for input too.

Is this missing the actual vector rendering though?

frontend/src/cljs/imas_seamap/core.cljs
JoshGx2000: [This kind of thing](https://stackoverflow.com/a/60472172) could probably be used to avoid polluting the reframe namespace, but we're still left with the side effects issue.

frontend/src/cljs/imas_seamap/views.cljs
olivergeorge: Hmm, I guess the options are:
* in component (analog of image GET which happens as side effect of rendering)
* in handler (normal place for making requests with possible benefit of allowing data to be held in app-db for reuse) 

Generally I lean away from complex views which need to process data or embed business logic used to massage data before use.

So... it depends but there seem to be clear options and factors to help pick between them.

What am I missing?

JoshGx2000: Resolved in Seamap chat 🙂

PR #117, opened 2022-09-01
Feature: (ISA-263) Richer main national layer card UI

JoshGx2000:
[ISA-263](https://jira.its.utas.edu.au/browse/ISA-263)

Messy diffs incoming!

On top of starting a basic UI for a richer main national layer card, a lot of the layer card component stuff has been refactored and should be much neater. Of most note is that the components were cleaned up with descriptions and moved into a separate view file.

As a part of the cleaning, unused components in the old view have been removed, and unused css classes have been removed (there could be selectors that I missed, but I was quite thorough).

Pretty pictures of main national layer card:
Screen Shot 2022-09-01 at 12 31 12 pm
Screen Shot 2022-09-01 at 12 31 47 pm
Screen Shot 2022-09-01 at 12 32 37 pm

The UI elements are "dumb" and don't actually affect the display of the layers in the map – separate ticket when we get to it.

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/layer_views.cljs
+++ b/frontend/src/cljs/imas_seamap/map/layer_views.cljs
@@ -94,3 +94,35 @@
    {:elevation 1
     :class     "layer-card"}
    [layer-card-content props]])
+
+(defn- layer-catalogue-controls [{:keys [layer] {:keys [active?]} :layer-state}]
+  [:div.layer-controls
+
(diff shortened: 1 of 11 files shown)


markhepburn (approved):
Not so much messy, as hard to follow!  (ie, what's new and what's just a separate-file refactor)

Looks fine, but probably easier in the future to do the refactor separately.

PR #119, opened 2022-09-05
(ISA-245) pip-audit github action

JoshGx2000:
[ISA-245](https://jira.its.utas.edu.au/browse/ISA-245)

I did some reading on [pip-audit](https://github.com/pypa/gh-action-pip-audit#configuration) and general [GitHub workflow actions syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) and managed to come up with the following file.

I think it should run Sunday 00:00 every week, and when it does it checks our `deploy_requirements.txt` for any vulnerabilities. We should be able to add vulnerabilities to ignore as needed.

In regards to receiving notifications if the workflow run fails, I believe that would be me [according to this](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs).

### DIFF
--- a/.github/workflows/auditrequirements.yml
+++ b/.github/workflows/auditrequirements.yml
@@ -1,0 +1,20 @@
+name: Audit requirements
+
+on:
+  schedule:
+    - cron:  '0 0 * * 0' # 0:00 every Sunday
+
+jobs:
+  pip-audit:
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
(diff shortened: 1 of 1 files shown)


markhepburn:
Nice research, thanks (so I could update the cron trigger and I'd receive notifications too, if I read correctly)
(Note: that's midnight UTC, so it's actually 10am here)

PR #128, opened 2022-09-15
Bugfix: (ISA-282) JSON legend fixes

JoshGx2000:
[ISA-282](https://jira.its.utas.edu.au/browse/ISA-282)

Point layer support, and using layer legend image as backup if unsupported by the app.

Screen Shot 2022-09-15 at 12 54 13 pm

I did consider using SVGs for the point layer legend keys, but that would take extra time (and I'm not sure how I'd dynamically set something like the stroke and colours based on the source data).

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -658,7 +658,10 @@
                      (fn [{:keys [title filter symbolizers]}]
                        {:label  title
                         :filter filter
-                        :color  (-> symbolizers first :Polygon :fill)})))]
+                        :style
+                        {:background-color (-> symbolizers first :Polygon :fill)
+                         :height           "100%"
+                         :width            "100%"}})))]
     (assoc-in db [:map :legends id] legend)))
 
 (defmethod get-layer-legend-success :feature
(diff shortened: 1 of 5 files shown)


frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -5,7 +5,7 @@
   (:require [clojure.string :as string]
             [re-frame.core :as re-frame]
             [cljs.spec.alpha :as s]
-            [imas-seamap.utils :refer [encode-state ids->layers first-where index-of]]
+            [imas-seamap.utils :refer [encode-state ids->layers first-where index-of append-query-params-from-map]]
```
markhepburn, line 8: We've possibly had this discussion, but that `append-query-params-from-map` could probably be simplified to `append-query-params` (most interfaces along these lines use maps/dicts, and there's no real value in specifying the type)

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -652,13 +652,40 @@
...
+
+                            (-> symbolizers first :Point)
+                            (let [{:keys [graphics size]} (-> symbolizers first :Point)
+                                  {:keys [mark fill stroke stroke-width]} (first graphics)]
+                              (merge
+                               {:background-color fill
+                                :border           (str "solid " stroke-width "px " stroke)
+                                :width            (str size "px")
+                                :height           (str size "px")}
+                               (when (= mark "circle") {:border-radius "100%"}))))}))))]
```
markhepburn, line 688: Just from he size of this diff I'm in favour of extracting that out into a separate function.  (not sure after that if it's any further benefit making it a `defmulti` or just extracting the code as-is; see what you think)

markhepburn (changes requested):
Looks good, couple (ok, one related, one unrelated) suggestions.

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Yep, absolutely right, fixed

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Done

PR #129, opened 2022-09-21
(ISA-265) ESRI GetFeatureInfo equivalent

JoshGx2000:
[ISA-265](https://jira.its.utas.edu.au/browse/ISA-265)

Did a little bit of tidying on the code that gets feature info, so now requests are dispatched in multimethods that depend on the layer info format type (so it should now be extendible to future format types should the need arise).

Can now also handle ESRI layers. It's very handy that L.esri.query returns a FeatureCollection response, since that's what we were already using for the JSON layers. ESRI feature layers responses display the same as the JSON layers.
Screen Shot 2022-09-21 at 3 05 48 pm

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/events.cljs
+++ b/frontend/src/cljs/imas_seamap/map/events.cljs
@@ -5,8 +5,8 @@
   (:require [clojure.string :as string]
             [re-frame.core :as re-frame]
             [cljs.spec.alpha :as s]
-            [imas-seamap.utils :refer [encode-state ids->layers first-where index-of append-query-params]]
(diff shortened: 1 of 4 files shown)


frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -46,82 +47,120 @@
...
-                    :VERSION       "1.1.1"}]
-        {:method          :get
-         :uri             server-url
-         :params          params
-         :response-format (ajax/text-response-format)
-         :on-success      [:map/got-featureinfo request-id point info-format]
-         :on-failure      [:map/got-featureinfo-err request-id point]}))))
-
-(defn get-feature-info [{:keys [db]} [_ {:keys [size bounds]} point]]
+(defmulti get-feature-info #(second %2))
```
markhepburn, line 50: Minor point, but I often like naming the arguments in the dispatch function, just to make it clear what arguments it's receiving.

markhepburn (approved):

frontend/src/cljs/imas_seamap/map/events.cljs
JoshGx2000: Ah cool, makes sense. Will do so in future 👍

PR #130, opened 2022-09-26
Fix some data types in SQL tables and procedures; add dependency; Fix a typo in email notification; Load config and write log by absolute path

khanh-condense:


etl/requirements.txt
```diff
@@ -1 +1,2 @@
 pyodbc==4.0.34
+pyproj==3.0.1
```
markhepburn, line 2: Just a couple of quick points here -- it's worth a comment to explain why this isn't the latest version, and also I've started checking the "pip freeze" output instead of the top level dependencies (with the rationale that this makes dependency checking more robust because you know exactly what you have installed)

markhepburn (approved):
I left a minor comment, when you get to it.  Thanks.

PR #132, opened 2022-09-28
(ISA-289) URL base init refactor

JoshGx2000:
[ISA-289](https://jira.its.utas.edu.au/browse/ISA-289)

URL db config has changed so that it's determined at boot based on the app page URL. If we want to specify additional base URL configs, all we have to do is add them to the app's default db config.

### DIFF
--- a/.github/workflows/buildfrontend.yml
+++ b/.github/workflows/buildfrontend.yml
@@ -82,6 +82,22 @@ jobs:
           app_tag=$(echo $app_tag | sed 's/^"//;s/"$//')
           sed -i "s/app.js/$app_tag/" 'resources/public/index.html'
 
+      - name: Conditionally update base URLs in index (master)
+        if: endswith(github.ref, '/master')
+        run: |
+          sed -i 's/(?<=api_url_base = ").+(?=";)/https:\/\/data.imas.utas.edu.au\/seamap\/api\//' 'resources/public/index.html'
+          sed -i 's/(?<=media_url_base = ").+(?=";)/https:\/\/data.imas.utas.edu.au\/seamap\/media\//' 'resources/public/index.html'
+          sed -i 's/(?<=wordpress_url_base = ").+(?=";)/https:\/\/seamapaustralia.org\//' 'resources/public/index.html'
+          sed -i 's/(?<=img_url_base = ").+(?=";)/\/app\/img\//' 'resources/public/index.html'
+
+      - name: Conditionally update base URLs in index (develop)
+        if: endswith(github.ref, '/develop')
+        run: |
+          sed -i 's/(?<=api_url_base = ").+(?=";)/https:\/\/data-dev.imas.utas.edu.au\/seamap\/api\//' 'resources/public/index.html'
+          sed -i 's/(?<=media_url_base = ").+(?=";)/https:\/\/data-dev.imas.utas.edu.au\/seamap\/media\//' 'resources/public/index.html'
+          sed -i 's/(?<=wordpress_url_base = ").+(?=";)/https:\/\/seamapaustralia-dev.imas.utas.edu.au\//' 'resources/public/index.html'
+          sed -i 's/(?<=img_url_base = ").+(?=";)/\/app\/img\//' 'resources/public/index.html'
+
       - run: tar -c -z -v -C resources/public -f "${{ env.GITHUB_REPOSITORY_SLUG }}-${{ env.GITHUB_REF_SLUG }}-frontend.tgz" .
 
       - name: Create timestamp for release tag-name
--- a/frontend/resources/public/index.html
+++ b/frontend/resources/public/index.html
@@ -18,9 +18,9 @@
     <script src="js/app.js"></script>
     <script>
       imas_seamap.db.api_url_base = "http://localhost:8000/api/";
-      imas_seamap.db.media_url_base = "http://localhost:8000/api/";
-      imas_seamap.db.wordpress_url_base = "https://seamapaustralia-dev.imas.utas.edu.au/";
-      imas_seamap.db.img_url_base = "/app/img/";
+      imas_seamap.db.media_url_base = "http://localhost:8000/media/";
+      imas_seamap.db.wordpress_url_base = "http://localhost:8888/";
+      imas_seamap.db.img_url_base = "/img/";
     </script>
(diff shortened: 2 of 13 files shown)


frontend/src/cljs/imas_seamap/db.cljs
```diff
@@ -91,21 +86,38 @@
...
+                                 "https://seamapaustralia.org/map/"
+                                 {:api-url-base       "https://data.imas.utas.edu.au/seamap/api/"
+                                  :media-url-base     "https://data.imas.utas.edu.au/seamap/media/"
+                                  :wordpress-url-base "https://seamapaustralia.org/"
+                                  :img-url-base       "/app/img/"}
+                                 "https://seamapaustralia-dev.imas.utas.edu.au/map"
+                                 {:api-url-base       "https://data-dev.imas.utas.edu.au/seamap/api/"
+                                  :media-url-base     "https://data-dev.imas.utas.edu.au/seamap/media/"
+                                  :wordpress-url-base "https://seamapaustralia-dev.imas.utas.edu.au/"
+                                  :img-url-base       "/app/img/"}}
```
markhepburn, line 121: (discussed offline, but leaving a paper-trail here).  This is nicely data-driven, but it still leaves the problem that for a new deployment we have to update the source.  I'd like these kind of external concerns to be configuration variables; no need for a new build, or a source change -- just update the external configuration (which in our case is probably the html page / javascript variables)

frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -105,15 +104,13 @@
         :map.layer.download/geotiff "GeoTIFF"}
        type-key))
 
-(defmulti download-link (fn [_layer _bounds download-type] (type->servertype download-type)))
+(defmulti download-link (fn [_layer _bounds download-type _api-url-base] (type->servertype download-type)))
```
markhepburn, line 107: Good catch

markhepburn (changes requested):
I think you're on the right path, it's just the "source" of the configuration that we need to change.

.github/workflows/buildfrontend.yml
```diff
@@ -82,6 +82,22 @@ jobs:
...
+          sed -i 's/(?<=img_url_base = ").+(?=";)/\/app\/img\//' 'resources/public/index.html'
+
+      - name: Conditionally update base URLs in index (develop)
+        if: endswith(github.ref, '/develop')
+        run: |
+          sed -i 's/(?<=api_url_base = ").+(?=";)/https:\/\/data-dev.imas.utas.edu.au\/seamap\/api\//' 'resources/public/index.html'
+          sed -i 's/(?<=media_url_base = ").+(?=";)/https:\/\/data-dev.imas.utas.edu.au\/seamap\/media\//' 'resources/public/index.html'
+          sed -i 's/(?<=wordpress_url_base = ").+(?=";)/https:\/\/seamapaustralia-dev.imas.utas.edu.au\//' 'resources/public/index.html'
+          sed -i 's/(?<=img_url_base = ").+(?=";)/\/app\/img\//' 'resources/public/index.html'
+
```
markhepburn, line 100: Aah -- thanks for that, but the goal is we don't need them here either (this is effectively still the issue that we have a deployment per branch, that we're trying to move away from).  We'll have basically this code still, but run at deploy-time via ansible, instead of github-actions.

frontend/src/cljs/imas_seamap/db.cljs
```diff
@@ -3,10 +3,10 @@
...
 (ns imas-seamap.db)
 
-(goog-define api-url-base "http://localhost:8000/api/")
-(goog-define media-url-base "http://localhost:8000/media/")
-(goog-define wordpress-url-base "http://localhost:8888/")
-(goog-define img-url-base "/img/")
+(def api-url-base       "http://localhost:8000/api/")
+(def media-url-base     "https://data-dev.imas.utas.edu.au/seamap/media/")
+(def wordpress-url-base "https://seamapaustralia-dev.imas.utas.edu.au/")
+(def img-url-base       "/app/img/")
```
markhepburn, line 9: Any reason these don't default to localhost now?  (just from a consistency/documentation standpoint)

frontend/resources/public/index.html
```diff
@@ -16,6 +16,12 @@
     <div id="toaster"></div>
     <script src="https://use.fontawesome.com/d25c01a5e8.js"></script>
     <script src="js/app.js"></script>
+    <script>
+      imas_seamap.db.api_url_base = "http://localhost:8000/api/";
+      imas_seamap.db.media_url_base = "http://localhost:8000/media/";
+      imas_seamap.db.wordpress_url_base = "http://localhost:8888/";
+      imas_seamap.db.img_url_base = "/img/";
+    </script>
```
markhepburn, line 24: Huh, I hadn't thought of doing it that way!  I think it should be safe (the risk of name-mangling always makes me nervous, but you can protect against that.. actually have you tested with advanced optimisations on?)

The monkey-punching aspect of it I'm not a fan of though.  How about something (roughly) like:
```javascript
<script>
var
  api_url_base = "http://localhost:8000/api/",
  media_url_base = "....";
imas_seamap.core.init(api_url_base, media_url_base,...);
</script>
```
Ie, seamap is almost just a library; you pass it configuration params as it starts up?

markhepburn (changes requested):
Right idea now I think, I'd just prefer a different execution (happy to discuss)

frontend/src/cljs/imas_seamap/db.cljs
JoshGx2000: Shoot, because I made a mistake, whoops. Fixed now.

frontend/resources/public/index.html
JoshGx2000: Fixed!

.github/workflows/buildfrontend.yml
JoshGx2000: Did you want me to leave it in for now, to be removed when the ansible deployment code has been updated, or has that all ready been taken care of?

markhepburn: Yeah, I guess leave it in thanks, then it's easily copyable for the playbook

markhepburn (approved):

PR #133, opened 2022-09-30
(ISA-263) Main national layer card functionality

JoshGx2000:
[ISA-263](https://jira.its.utas.edu.au/browse/ISA-263)

This feature is currently incomplete, and this PR is merely here to serve as a code review.

Some work has been done on the main national layer card timeline and alternate views sections, which is intended to modify how the main national layer appears on the map by substituting it with other layers specified in the DB.

Currently the national layer is replaced in the layers sub, but maybe there's a better way to handle the national layer as a special case so the code doesn't get confusing. Any ideas?

**Edit:** I _might_ have a neat way to solve the issues, but regardless you should read through the code as it is to see if anything sticks-out at you.

### DIFF
--- a/frontend/src/cljs/imas_seamap/map/subs.cljs
+++ b/frontend/src/cljs/imas_seamap/map/subs.cljs
@@ -45,7 +45,7 @@
               0.4)))))       ; Might be nice to make this configurable eventually
 
 (defn map-layers [{:keys [layer-state filters sorting]
-                   {:keys [layers active-layers hidden-layers bounds categories national-layer-timeline]} :map
+                   {:keys [layers active-layers hidden-layers bounds categories national-layer-timeline keyed-layers]} :map
                    :as _db} _]
   (let [all-layers      layers
         categories      (map-on-key categories :name)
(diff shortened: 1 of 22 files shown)


JoshGx2000:
I think I'm just going to separately layer state (opacity, loading, etc.) for the main national layer, separately from the other layers because it's a special case. That's the only way I can think of doing this in a neat way.

frontend/src/cljs/imas_seamap/map/subs.cljs
```diff
@@ -78,6 +82,20 @@
   (let [categories (get-in db [:map :categories])]
     (map-on-key categories :name)))
 
+(defn national-layer-year [db _]
+  (:year
+   (or ; we use 'or' here instead of having an additional param in 'get-in' because get-in will actually read nil as a success and return that
```
markhepburn, line 87: Makes sense (common-lisp has two return values; the value itself, and if the key was found -- which gets around this issue).  Might be worth thinking about if the key should be present if it's nil, I guess.

markhepburn (commented):
I see what you mean.  Am I right that the national-layer specific stuff in the layers sub is for configuring the UI of the richer main-layer card?  Might it make sense to pull that into a separate sub?  (I'm not sure where that sub gets invoked, it just feels cleaner: the sub takes care of returning the appropriate layer for rendering, whether it's a different view of the main layer or just a regular layer, and that part of the code doesn't need to care.  But as I said, where the richer card gets initialised from is less clear.  This is probably one of those cases that might suddenly make the TMA-support refactor tricky, by the way!

JoshGx2000:
Layer state should be appropriately shared between the different displayed national layers.
Screen Shot 2022-10-03 at 11 13 14 am

markhepburn (approved):

PR #141, opened 2022-10-19
(ISA-192) Region reports research effort graphs

JoshGx2000:
[ISA-192](https://jira.its.utas.edu.au/browse/ISA-192)

Adaptive research effort graphs for the region report data pages. Data for the graphs is queried and stored in the Django server media storage, where it is then retrieved via AJAX and displayed as a graph using Vega.

You cannot argue with the results:
Screen Shot 2022-10-19 at 3 38 35 pm

### DIFF
--- a/wordpress/plugins/region-report/single-region_report.php
+++ b/wordpress/plugins/region-report/single-region_report.php
@@ -4,6 +4,7 @@
     $habitat_statistics_url = get_post_meta(get_the_ID(), 'habitat_statistics_url', true);
     $bathymetry_statistics_url = get_post_meta(get_the_ID(), 'bathymetry_statistics_url', true);
     $habitat_observations_url = get_post_meta(get_the_ID(), 'habitat_observations_url', true);
+    $research_effort_url = get_post_meta(get_the_ID(), 'research_effort_url', true);
     $region_report_data_url = get_post_meta(get_the_ID(), 'region_report_data_url', true);
 ?>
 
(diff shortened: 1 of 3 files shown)


wordpress/plugins/region-report/single-region_report.php
```diff
@@ -368,7 +370,137 @@ function addToList(list, text) {
...
+                    <div id="region-report-research-effort-<?php the_ID(); ?>-2"></div>
+                </div>
+                <script>
+                     postElement.addEventListener(
+                        "researchEffort",
+                        e => {
+                            const values = [];
+                            e.detail.forEach(e => {
+                                values.push({
+                                    year: e.year,
```
markhepburn, line 385: You have two distinct bindings for "e" here, which might get confusing (sometimes it's a useful "trick" that makes code read clearer, but I worry here that it's hard to keep track of which e is what)

markhepburn (approved):
Minor comment, otherwise lgtm.  I know we briefly raised earlier whether it mightn't make as much sense to use the API endpoint, but I don't know that it's important (unless I'm forgetting something the only benefit is being able to control CORS headers ourselves, but that's a once-off fix anyway)

JoshGx2000:
@markhepburn Will switch over to API endpoint in other cases, however for the case of the research effort graphs there's an issue with how long the queries take that makes caching the results in the media storage a more effective solution.

markhepburn:
(just coming back to this, reviewing the last meeting notes) I think the main concern was ensuring that the data is the same between sidebar and report pages, which presumably means the API is the same where applicable.

PR #142, opened 2022-10-19
(ISA-178) Add 5% decrease check; fix a bug in IMPORT_MARS

khanh-condense:


markhepburn (commented):
It broadly looks fine; I'm assuming you've tested it, but the other thing I'd be looking at is the int/float conversion and if there's any boundary conditions there.

I'll leave this open for now; Peter and Emma raised a couple of things so let's go through the entire thing when you're back on deck.

markhepburn:
(while I think of it, the preferred branch naming is "feature/<ticketnum>-description", eg "feature/isa178-seamap-etl")

markhepburn:
@khanh-condense when you're back let's have a code walk-through; there's a couple of python things I picked up, and it's worth going over the SQL together too I think.

PR #144, opened 2022-11-01
(ISA-243) Region Report Pressures

JoshGx2000:
[ISA-243](https://jira.its.utas.edu.au/browse/ISA-243)

New widget added to region report pages showing categorised pressure maps for a region:
Screen Shot 2022-11-01 at 6 02 20 pm

A new Django management command generates the pressure preview images, much in the same way the layer preview images are generated in the Seamap app.

### DIFF
--- a/backend/catalogue/management/commands/generate_pressure_previews.py
+++ b/backend/catalogue/management/commands/generate_pressure_previews.py
@@ -8,7 +8,7 @@ from PIL import Image
 from io import BytesIO
 import logging
 
-from catalogue.models import KeyedLayer, RegionReport
+from catalogue.models import KeyedLayer, RegionReport, Pressure
 
 Image.MAX_IMAGE_PIXELS = None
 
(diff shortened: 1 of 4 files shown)


backend/catalogue/management/commands/generate_pressure_previews.py
```diff
@@ -0,0 +1,161 @@
+from django.core.management.base import BaseCommand
+from django.core.files.storage import default_storage
+from django.core.files import File
+from django.conf import settings
+from urllib.request import urlopen
+from urllib.parse import urlencode
+from PIL import Image
```
markhepburn, line 7: Trivial thing, but I think pep8 groups (sorted) imports into built-ins/third-party/local.

backend/catalogue/management/commands/generate_pressure_previews.py
```diff
@@ -0,0 +1,161 @@
...
+        'layers': layer.layer_name,
+        'styles': layer.style or '',
+        'format': 'image/png',
+        'transparent': 'true',
+        'version': '1.1.1',
+        'width': size['width'],
+        'height': size['height'],
+        'srs': 'EPSG:4326',
+        'bbox': '{west},{south},{east},{north}'.format(**bbox),
+        'cql_filter': f'RESNAME=\'{park}\'' if park else f'NETNAME=\'{network}\''
```
markhepburn, line 76: Minor: you can use f-strings with double-quotes here, which would avoid the need to escape the internal single-quotes.  Might look out of place with the rest of the entries though?  (not sure what the single/double preference is in pep8 actually; I assume they have opinions)

markhepburn (approved):

PR #150, opened 2022-11-09
(ISA-306) Style Tweaks

JoshGx2000:
[ISA-306](https://jira.its.utas.edu.au/browse/ISA-306)

A few style and formatting tweaks. This is not an implementation of the styles outlined in the Nat document (though it does include the mentioned fonts).


frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -238,8 +238,8 @@
             :transform  (str "translate3d(" x "px, " y "px, 0px)")
             :z-index    700}}
    (if (> distance 1000)
-     (gstring/format "%.2f km" (/ distance 1000))
-     (gstring/format "%.0f m" distance))])
+     (str (format-number (/ distance 1000) 2) " km")
+     (str (format-number distance 0) " m"))])
```
markhepburn, line 242: Haven't actually noticed these before; convention is usually for the unit to appear after without a space isn't it?  ("12km" rather than "12 km").  Not complaining and clearly I haven't noticed before.

markhepburn (approved):

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Fixed, thanks

PR #220, opened 2024-07-12
Feature: Carbon abatement enhancements

JoshGx2000:


.github/workflows/buildfrontend.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           echo "PROFILE=min" >> $GITHUB_ENV
       - name: Conditionally set build profile based on branch (develop)
-        if: endswith(github.ref, '/develop')
+        if: ${{ !endswith(github.ref, '/master') }}
```
markhepburn, line 34: Huh, I would not have guessed that

backend/catalogue/admin.py
```diff
@@ -57,3 +63,34 @@ class RegionReportAdmin(admin.ModelAdmin):
         ('minx','maxx','miny','maxy',),
     )
 admin.site.register(models.RegionReport, RegionReportAdmin)
+
```
markhepburn, line 66: Random minor one, but the general python spacing I try and stick with is "pep8", which is 2 blank lines between top-level classes/functions, and a single blank line in between eg class methods, or for readability.

backend/catalogue/models.py
```diff
@@ -114,7 +114,50 @@ def server_info(self):
...
+    def cql_property_values(self, cql_properties: list) -> dict:
+        params = {
+            'service': 'WFS',
+            'version': '2.0.0',
+            'request': 'GetFeature',
+            'typeNames': self.layer_name,
+            'outputFormat': 'application/json',
+            'propertyName': f"({','.join(cql_properties)})",
+        }
+        r = requests.get(url=self.server_url, params=params, verify=False)
```
markhepburn, line 127: I'm trying to decide how I feel about making http requests from a model method.  It's a bit "unexpected" (db queries is obviously standard; http might not be).  *this is  not thought through:* you could perhaps separate out the query and the processing; maybe a separate method to create the query and this one to parse it, or perhaps have this method take the "http queryer" that's then used by this method (basically, passing in requests, but it could be a bit abstracted.  It makes it clear to the caller that it'll be doing a network request).  Both of those options would make it simpler to test, as well, which is often a useful sniff-test.

frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -416,34 +416,131 @@
...
+        alternate-view-rich-layer-id (get-in db [:map :rich-layers :layer-lookup alternate-views-selected-id])
+        alternate-view-rich-layer    (first-where #(= (:id %) alternate-view-rich-layer-id) (get-in db [:map :rich-layers :rich-layers]))
+
+        timeline                  (mapv #(->timeline % db) (or (:timeline alternate-view-rich-layer) timeline))
+        timeline-selected         (first-where #(= (get-in % [:layer :id]) timeline-selected-id) timeline)
         slider-label              (or (:slider-label alternate-view-rich-layer) slider-label)
-        timeline-selected         (first-where #(= (get-in % [:layer :id]) timeline-selected-id) timeline)]
+
+        controls                  (mapv #(->control % rich-layer db) controls)
+        cql-filter                (apply
```
markhepburn, line 495: This sort of thing inline definitely makes me think we could/should have some CQL utilities :)

markhepburn (approved):
Comments were mostly higher-level and musings, but I think an overarching impression (and I'm just as much to blame) is the size of Seamap is going to make it more important to think about clarity and making it easier to navigate.  That might mean more utility modules, more comments, etc.

frontend/src/cljs/imas_seamap/map/utils.cljs
JoshGx2000: Yep, exactly. Working on it now, actually (since I need them for the smart pill region selection too now)

JoshGx2000:
> Comments were mostly higher-level and musings, but I think an overarching impression (and I'm just as much to blame) is the size of Seamap is going to make it more important to think about clarity and making it easier to navigate. That might mean more utility modules, more comments, etc.

Honestly, I agree, but I'm not sure when to do that. I would really like to restructure stuff, but it's hard to justify when that sort of cleaning can break the end app.

backend/catalogue/models.py
```diff
@@ -114,7 +144,417 @@ def server_info(self):
...
+            if 'fill' in graphic:
+                style['backgroundColor'] = graphic['fill']
+
+            # Set border to stroke width and color if present
+            if ('stroke-width' in graphic) or ('stroke' in graphic):
+                style['border'] = f"{graphic.get('stroke-width', 2)}px solid {graphic.get('stroke', 'black')}"
+
+            # Custom style for circle markers
+            if 'mark' in graphic and graphic['mark'] == 'circle':
+                style['borderRadius'] = '100%'
```
markhepburn, line 427: This a general thought, in the context of that "Philosophy of Software Design" book I mentioned last team meeting.  (and I do like the docstrings and the generally-improved documentation!).  One of things mentioned, obvious once you've heard it, is "comments that just repeat the code aren't useful".  So, this is arguably a case of that.  What might be more useful is a bit of context that isn't immediately obvious from the following lines; eg instead of "set border..." it might be a note that a border can be defined in one of two ways so we need to handle both... that sort of thing.

(I think it's super hard to teach commenting to first-years, because they don't have the context of what is unexpected, and you just get a bunch "i++; // add one to i")

JoshGx2000: I tried some improvement (here at least) with b7ab443

backend/catalogue/admin.py
JoshGx2000: Understood, I will use pep8 as my guide from now om

PR #221, opened 2024-07-12
Feature: Region reports enhancements

JoshGx2000:


backend/catalogue/management/commands/retrieve_squidle_annotations_data.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--skip_existing',
+            help='If true, skips retrieving squidle annotations data where we already have it',
+        )
+
+
+    def handle(self, *args, **options):
```
markhepburn, line 90: General comment on this -- it's getting quite long-winded and nested, so I think at the very least a few comments about what it's doing, and the high-level steps along the way would be beneficial.

backend/catalogue/models.py
```diff
@@ -150,7 +150,7 @@ class SaveState(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     hashstate = models.CharField(max_length = 8000)
     description = models.TextField(blank=True, null=True)
-    time_created = models.DateTimeField(default=datetime.now())
+    time_created = models.DateTimeField(default=django.utils.timezone.now)
```
markhepburn, line 153: Hmm, I hadn't noticed that before; doesn't really matter here, but any reason that was explicitly a local-time?  (I'm used to defaulting to UTC in databases, unless there's a compelling reason otherwise)

backend/habitat/viewsets.py
```diff
@@ -1328,17 +1339,90 @@ def region_report_data(request):
...
+        ).prefetch_related(
+            'category',
+            'data_classification',
+            'organisation',
+            'server_type'
+            ),
+        many=True
+    ).data
+
+    data["all_layers"] = list(reversed([next(layer for layer in layers if layer["id"] == keyed_layer['layer']) for keyed_layer in keyed_layers if keyed_layer['keyword'] == 'data-report-minimap-panel1']))
```
markhepburn, line 1372: This sort of block in particular could definitely do with some commenting

backend/habitat/viewsets.py
```diff
@@ -1328,17 +1339,90 @@ def region_report_data(request):
...
+        'outputFormat': 'application/json',
+        'cql_filter': (f"RESNAME='{park}'" if park else f"NETNAME='{network}'")
+    }
 
-    with connections['transects'].cursor() as cursor:
-        cursor.execute(SQL_GET_PARK_SQUIDLE_URL if park else SQL_GET_NETWORK_SQUIDLE_URL, [park or network])
-        entry = cursor.fetchone()
-        data["squidle_url"] = entry[0] if entry else None # some park
+    try:
+        r = http_session().get(url=boundary_simplified['server_url'], params=params)
```
markhepburn, line 1393: This is the sort of thing that jumps out when you're reviewing; making an http call from another view (we're not doing microservices, so we don't have that kind of infrastructure).  Apart from the general inlining, I wonder if this wouldn't be better extracted into a smaller method to get the boundaries, and then that could potentially be cached?  (was this what we were talking about django's caching infrastructure for?  I can't remember)

markhepburn (approved):
So I ended up leaving a few comments on this one -- mostly reactions as much as concrete suggestions, but the general theme I think is just that the viewsets is getting quite long!  That's because we've gradually accumulated functionality, and the main implication I think is that it can be hard to follow; that could be improved with a little bit of refactoring and commenting.  A whole-file refactoring (eg into smaller apps) is probably not on the cards right now, but it's worth keeping on mind as we keep adding to it.

wordpress/plugins/region-report/region-report.js
```diff
@@ -609,10 +749,35 @@ class RegionReport {
...
+                layers = layers.concat(this.allLayers.metadata?.layers);
+                break;
+            case 'public':
+                layers = layers.concat(this.publicLayers.metadata?.layers);
+                break;
+        }
+
         return [
             "^ ",
             "~:autosave?",
```
markhepburn, line 783: Jeebus, is this manually emitting transit-js?

JoshGx2000: Umm, yes.
Sorry, it was the only solution I could come up with at the time, and I was clearly too lazy to make a utility of JSON to do the conversion
Now that I know what it is, I should use a package to do that.

backend/catalogue/management/commands/retrieve_squidle_annotations_data.py
```diff
@@ -0,0 +1,211 @@
...
+        amp_depth_zones = AmpDepthZones.objects.all().values('netname', 'resname', 'zonename', 'min', 'max').distinct()
+        network_region_depth_zones = {}
+        for amp_depth_zone in amp_depth_zones:
+            if amp_depth_zone['netname'] not in network_region_depth_zones:
+                network_region_depth_zones[amp_depth_zone['netname']] = []
+            network_region_depth_zones[amp_depth_zone['netname']].append({k: amp_depth_zone[k] for k in ['resname', 'zonename', 'min', 'max']})
+        region_depth_zones = {}
+        for network, depth_zones in network_region_depth_zones.items():
+            if network not in region_depth_zones:
+                region_depth_zones[network] = {}
```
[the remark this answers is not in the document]
JoshGx2000: Wow, thanks. I will keep that in mind (I should have used that a few times!)

backend/catalogue/management/commands/retrieve_squidle_annotations_data.py
JoshGx2000: Reworked it a bit to be a bit less confusing...

backend/catalogue/models.py
JoshGx2000: Yeah, this isn't really that important. I only changed it because the console was giving a warning for the field every time I started the server.

PR #294, opened 2026-02-10
Experiment: configuration-driven deployments

markhepburn:
**DO NOT MERGE YET!!**

This is an experiment for feedback.  (Claude wrote most of the code, there's a few dangling ends that could be tidied up)

*Motivation:* We have a growing number of deployments all sharing the same codebase.  We accomodate a few tweaks between the different versions, but the methodology (a separate entrypoint, with different UI components overridden as needed) is starting to feel a bit strained.

*Goal:* a single entrypoint, with a configuration being chosen (currently by an arg to `init` to select from some pre-defined configurations, but it could be sourced from an API call, etc).

*Approach:* The goal is a fairly light touch, with feature-flags used where necessary, rather than a fully-dynamic system where entirely new UIs can be constructed.  So, key features are defined as keywords, which feed into a feature-flag subscription.  A few smaller pieces, such as log and spash-screen text, and a CSS class to specialise styling, are also configuration entries.  (also, map initial zoom/extent and CRS)

*TODO:* I haven't added anything yet to customise subs/handlers registration.  My current thinking is to have another registry, indexed by feature-flag again, where each the features configured for a given deployment are used to load a register a bag of subs and handlers.

### DIFF
--- a/frontend/docs/configuration-refactor-remaining-tasks.md
+++ b/frontend/docs/configuration-refactor-remaining-tasks.md
@@ -1,0 +1,443 @@
+# Configuration-Driven Refactor: Remaining Tasks
+
+## Status Summary
+
+### ✅ Completed Phases
+
+- **Phase 1: Extract Configuration** - All deployment configs, content registry created
+- **Phase 2: Feature Subscriptions** - Feature flags and branding subscriptions implemented
+- **Phase 3: Migrate Leaf Components** - Eliminated `tma?` parameter from all leaf components
+- **Test Harness** - Automated tests and deployment switcher UI created
+
+### 🔄 Current Status
+
+Ready for manual testing of the configuration system using the deployment switcher.
+
+### 📋 Remaining Work
+
+1. **Manual Testing** (Next immediate step)
+2. **Phase 4: Unify Layouts**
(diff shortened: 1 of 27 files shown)


frontend/docs/configuration-refactor-remaining-tasks.md
```diff
@@ -0,0 +1,443 @@
...
+  (:require [imas-seamap.config.deployments :as deployments]))
+
+(defn detect-deployment []
+  "Detect which deployment to use based on URL or environment"
+  (let [hostname (.-hostname js/location)]
+    (cond
+      (re-find #"tasmarineatlas" hostname)    :tas-marine-atlas
+      (re-find #"antarctica" hostname)        :seamap-antarctica
+      (re-find #"futuresofseafood" hostname)  :futures-of-seafood
+      :else                                   :seamap-australia)))
```
JoshGx2000: Maybe this got buried, but I think the idea of "deployment detection" based on hostname is a bad idea.

JoshGx2000: Ah, I missed that "How should production deployment detect which config to use?" is a question still to be resolved, and that this is one possible approach 👍

PR #296, opened 2026-02-13
ISA-690: WIP WMS timeseries support based on leaflet-timedimension

JoshGx2000:
[ISA-690](https://utas.atlassian.net/browse/ISA-690)

### Friction in adding timeseries support
Documenting the list of issues that occurred when trying to add timeseries support through `leaflet-timedimension`.
- Importing `leaflet-timedimension` breaks existing code because it causes a second instance of Leaflet to be loaded into the module graph. Leaflet and its plugins don’t follow the rules of ES6-style imports, and plugins (e.g. `leaflet-timedimension`) end up extending different `L` objects, which hides extensions added by other plugins (e.g. `leaflet-easyprint`). 
  - This needs to be documented to save headaches in the future!
  - Vendoring `leaflet-timedimension` and modifying its code to avoid redefining the global `L` attached to the window prevents it from breaking existing code.
- Adapting  `L.timeDimension` and `L.control.timeDimension` into React Leaflet components through `ReactLeafletCore/createControlComponent` doesn’t work; using either inside the map component results on the app crashing with messages about an error occurring in a ForwardRef component. 
  - This seems like it will be a pain to debug, because the actual error is obscured. I believe the ForwardRef mentions come from `ReactLeafletCore/createControlComponent`, and the ForwardRef to which it is referring to is a ref to the map.

### What’s been done so far
- Vendored and modified `leaflet-timedimension` to avoid redefining the global `L` attached to the window, to prevent it from breaking existing code.
- Added a warning to `leaflet.cljs` to inform developers about the particulars and complications of Leaflet imports.
- Added `wms-timeseries` layer component using `L.timeDimension.layer.wms`. Appears to function identically to a standard WMS layer right now (i.e. displays fine on map, but doesn’t have timeseries functionality).
- WIP adapting `L.timeDimension` and `L.control.timeDimension`. Usage crashes app.


frontend/src/cljs/imas_seamap/futuresofseafood/core.cljs
```diff
@@ -131,6 +131,7 @@
     :map/get-feature-info-map-server-step-2 [mevents/get-feature-info-map-server-step-2] ; MapServer layers need to make an additional request to determine if they are a group layer
     :map/got-featureinfo                  mevents/got-feature-info
     :map/got-featureinfo-err              mevents/got-feature-info-error
+    :map.time/info                        mevents/time-set-current-time
```
markhepburn: The name (`info`) stands out a bit here (it reads like a sub not a handler, for a start). Can we think of a more descriptive name?

JoshGx2000:
Running into some friction with the plugin and our setup @markhepburn. I can continue trying to fight it, but it might be a waste of time.

**Problem A**
This is moreso a problem that currently exists in Seamap causing issues. Every time the map-component containing the `time-dimension-control` component from our interop is updated, the `time-dimension-control` gets unmounted and remounted. This causes minor issues when the "play" option is used with the control, as the time dimension will continue playing through the timeseries layers even when the control is remounted, but the play/pause button will reset, leaving the user unable to pause the animation.
We could sidestep this by ignoring `time-dimension-control` and going with our own control, which we do intend to do with nathaz at some point, but it is annoying that the out-of-the-box stuff isn't working presently.

**Problem B**
We want to keep the current time the user has selected when they refresh the page or when they share a URL with someone else, but `leaflet-timedimension` doesn't play well with this.
When creating the time dimension component, I've tried to keep the current time we have stored in the re-frame DB by including it in the args:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/src/cljs/imas_seamap/futuresofseafood/map/views.cljs#L133-L135
It does work for setting the time initially, but then the time is changed once the layer capabilities are loaded. This is an intentional feature in leaflet-timedimension:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L999-L1004
The only path forward I see here is to cut out this chunk of code, or at least modify so it only works if there isn't a time already set.

I can see that I'm fighting the tools to get our desired behaviour, so I want to know if you think I'm approaching it wrong, @markhepburn.

markhepburn:
Shooting from the hip a little, but for B does it help if you change setDefaultTime?

I'd need for context for A.  Why is it remounting?

JoshGx2000:
No, changing setDefaultTime doesn't work on B. I was so confused about this, because the default value of setDefaultTime is supposed to be `false`!
I did a little bit of logging and found that was being set to `true` on this line:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L1000
I did experiment by changing the `||` to a `??` and got some slightly different behaviour with setting the initial times, but it still doesn't fully work as expected. I'll need a little more time to investigate.

For A, I'm not sure _why_ it's remounting, but I know that this section of code is constantly rerunning: https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/src/cljs/imas_seamap/interop/leaflet.cljs#L275-L279

JoshGx2000:
@markhepburn
`leaflet-timedimension` has been annoying to work with, but I've made a modification to fix problem B. I'll walk through the problem and the modification I've made:
We're having `leaflet-timedimension` request the time capabilities from the layer server so that we don't have to make that request ourselves, this is fine. Because of this, the `_requestTimeDimensionFromCapabilities` function is run from the WMS timeseries layer. That's this bit of code here:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L986-L1010

The line that's causing issues (but also helps, in a way) is `this._timeDimension.setCurrentTime(this._defaultTime);` on line 1003. This is supposed to set the `timeDimension`'s current time to the default time of the layer, supplied by the layer capabilities on the layer server. This is actually _good_ and I see the library author's intent behind the line! Without this line, the `timeDimension`'s `currentTime` will have a value of `null`. For an explanation of why _that_ is the case, I've left some extra reading at the bottom of this comment.

The line that sets the current time will run no matter what value for `setDefaultTime` is provided, because of this line here:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L1000

Because we haven't provided any times to our `timeDimension`, fully expecting them to come from the layer server capabilities, `this._timeDimension && this._timeDimension.getAvailableTimes().length == 0` will always evaluate to `true` .

Anyway, knowing that the current time of our `timeDimension` will always be set so long as we're getting our time dimension times from the layer server, we can use this part of the code to our advantage to update the value of `currentTime` the the value we _actually_ want (the one we're trying to hydrate the app with from our re-frame state). With a new option added to `timeDimension` for `defaultTime`, we can access that value here and set that as the current time for the `timeDimension`:
https://github.com/IMASau/Seamap/pull/296/changes/9c48363358498b944239da92880ffad0b1fa265c#diff-6757500c3936a611d8497c4bd7821d247d3bd7bb3c2a4e95e7fa9c2ab8d39a03L986-L1010

<br>

### Extra Reading: Why is timeDimension's currentTime null? 
When I first tried removing the `this._timeDimension.setCurrentTime(this._defaultTime);` line from the timeseries layer, I noticed that `timeDimension` had a value of `currentTime` of `null`, despite the fact that I'd supplied a `currentTime` in the options, and that it's one of the valid options: https://github.com/socib/Leaflet.TimeDimension#options

It turns out though, that supplying a `currentTime` in the options for `timeDimension` isn't good enough. The way the `currentTime` option works means it'll only set the current time if there is also a list of available times to select from, shown here in the constructor for `timeDimension`:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L43-L45

I see the library author's intent behind this! It doesn't make sense to set the current time of the `timeDimension` without having available times to select from. This is because of the way `setCurrentTime` works. The `timeDimension` object doesn't _actually_ store the current time, but instead the index of the current time from the list of available times! The `setCurrentTime` function takes the supplied time and finds the the index of the available times with the closest time, and sets that index to be the current time index:
https://github.com/IMASau/Seamap/blob/07b79b28fb30c2e5d6377b2cbaac438a6420630d/frontend/custom_modules/leaflet-timedimension/dist/leaflet.timedimension.src.js#L135-L138

If there aren't any available times, then the current time can't be set. This is the reason why `timeDimension`'s time is `null` before the timeseries layer has requested time dimension from capabilities.

I want to argue on the author's behalf here: what we are attempting to do is a bit of an unusual use-case. We want to set the current time on the map _before_ we even know if our layers we want to show on the map can use that time. That's a bit of a strange thing to do, so I can understand why it's not something that works well with how the library is designed.

### What now?
I'm confident that with this new code that problem B is resolved. Problem A remains an issue, though not a significant one. To my knowledge problem A is only impacting the play/pause functionality of the time control, and we'll be rolling our own control for the time dimension soon enough to where this won't be our concern.

I'm happy for a final review and merge now, @markhepburn.

JoshGx2000:
I'm just in the middle of my own code review, and I'm looking at this patch code you wrote @markhepburn
https://github.com/IMASau/Seamap/commit/fb3f0b34dae2ab98aab9f5ce058e2809f0a65137

My understanding from the commit description and the code is that it's supposed to be a patch that re-adds the functionalities that `leaflet-timedimension` adds to L.TileLayer, because our plugin import setup wipes those. Is that correct?

markhepburn:
> I'm just in the middle of my own code review, and I'm looking at this patch code you wrote @markhepburn [fb3f0b3](https://github.com/IMASau/Seamap/commit/fb3f0b34dae2ab98aab9f5ce058e2809f0a65137)
> 
> My understanding from the commit description and the code is that it's supposed to be a patch that re-adds the functionalities that `leaflet-timedimension` adds to L.TileLayer, because our plugin import setup wipes those. Is that correct?

If I'm understanding what you're saying, yes: that code is largely taken from the timedimension code itself.  From memory the issue was that the library monkey-patches Leaflet itself so it has a consistent interface to work with; my initial attempt just bolted those methods onto the timedimension layer, but the problem is when you animate things it is swapping out with new layers all the time, and they _won't_ have the new methods.  So, you need to patch `prototype` for them to be persistent.  It's pretty nasty.

frontend/src/cljs/imas_seamap/map/events.cljs
```diff
@@ -1122,3 +1150,34 @@
...
+    ;; * it has a timeDimension component, and
+    ;; * the time has changed from what's currently set in timeDimension,
+    ;; then do the side-effect of setting the time for the component.
+    (when time-dimension-ref
+      (let [time-dimension-current-time (.getCurrentTime time-dimension-ref)
+            time-changed?               (not= current-time time-dimension-current-time)]
+        (when time-changed?
+          (.setCurrentTime time-dimension-ref current-time))))
+    {:db (assoc-in db [:display :current-time] current-time)
+     :dispatch [:maybe-autosave]}))
```
markhepburn, line 1174: Should this be inside the `when`?

frontend/src/cljs/imas_seamap/map/utils.cljs
```diff
@@ -62,6 +73,18 @@
 
 (defn habitat-layer? [layer] (-> layer :category (= :habitat)))
 
+(defn has-time-dimension? [layer] (#{:wms-timeseries} (:layer_type layer)))
+
+(defn datetime-epoch-milliseconds-to-iso
```
markhepburn: It's pedantic, but the name of this is possibly overly descriptive (it's describing the literal implementation, not the purpose, semantics, etc).  Anything wrong with `timestamp-to-iso`, for eg?  (or similar).  Or `timestamp-ms-to-iso` if you wanted a bit of a type-hint in the name.  Seconds or millisecs since epoch is pretty standard, this makes it sound like it's some corner-case that needs to be explicitly called out.

markhepburn:
> I see the library author's intent behind this! It doesn't make sense to set the current time of the `timeDimension` without having available times to select from. This is because of the way `setCurrentTime` works. The `timeDimension` object doesn't _actually_ store the current time, but instead the index of the current time from the list of available times! The `setCurrentTime` function takes the supplied time and finds the the index of the available times with the closest time, and sets that index to be the current time index:

I just want to check (I haven't looked in the lib); what happens if you specify a singleton list of current time, can you still force it to fetch the available times?

```clojure
[:f> timeseries-layer {:current-time hydrated-current-time 
                       :available-layers  [hydrated-current-time] 
                       :update-times true
                       ...}]
```

markhepburn (approved):

JoshGx2000:
> I just want to check (I haven't looked in the lib); what happens if you specify a singleton list of current time, can you still force it to fetch the available times?

God, why didn't I think of that 🤦
I'll try that tomorrow morning. It sounds promising.

PR #319, opened 2026-06-16
NATHAZ-61: Side-By-Side Map Prototype

JoshGx2000:
[NATHAZ-61](https://utas.atlassian.net/browse/NATHAZ-61)

This is a simple side-by-side map prototype for Natural Hazards Atlas.

It uses the [Leaflet.Sync](https://github.com/jieter/Leaflet.Sync) library for functionality.

The purpose of this work was to demonstrate that side-by-side maps were viable, and to discover potential difficulties in a full adoption.

Includes rewrites to `map-component` component across the different Seamap apps. This was to reduce the number of unnecessary re-renders of the map, which was causing issues with the Sync library (because the actual Leaflet map ref in the code was constantly changing). _Side note: I'm quite happy that this feature resulted in a net negative of monkey code._

The feature is activated using a checkbox on the "current view" tab labeled "compare maps". This is currently only an option in the Natural Hazards Atlas app.

As a demonstration on how the maps can show different data on each, the right map will always show SSP2 for hazard layers:


frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -394,127 +377,121 @@
...
+          [:<>
+           (if (:side-by-side-views-selected rich-layer)
+             [side-by-side-layer
+              {:layer           layer
+               :boundary-filter boundary-filter
+               :layer-opacities layer-opacities
+               :cql-filter-fn   cql-filter-fn
+               :z-index         z-index
+               :rich-layer-fn   rich-layer-fn}]
+             [leaflet/pane {:name (str (random-uuid) (.now js/Date)) :style {:z-index z-index}}
```
markhepburn, line 428: Not part of this PR I get but the random+date gives me the hives every time I see it.  Is it possible to make layer-component a generic function and only wrap in a pane -- I can't remember why it has to be random -- when necessary?

frontend/src/cljs/imas_seamap/map/views.cljs
```diff
@@ -394,127 +377,121 @@
...
+                :layer-opacities layer-opacities
+                :cql-filter      (cql-filter-fn layer)}]])]))
+      visible-layers)]))
+
+(defn map-component []
+  (let [{:keys [center zoom bounds]}                @(re-frame/subscribe [:map/props])
+        feature-info                                @(re-frame/subscribe [:map.feature/info])
+        {:keys [query mouse-loc] :as transect-info} @(re-frame/subscribe [:transect/info])
+        {:keys [region] :as region-info}            @(re-frame/subscribe [:map.layer.selection/info])]
+    (js/console.log "Rerender map-component")
```
markhepburn: Useful debugging, but doesn't need to remain obviously

frontend/src/cljs/imas_seamap/natural_hazards_atlas/map/views.cljs
```diff
@@ -7,142 +7,191 @@
...
+    (r/track! #(when (and @map-a @map-b) (.sync @map-a @map-b) (.sync @map-b @map-a)))
+    (r/track!
+     (fn []
+       @split-ratio          ; re-render when split ratio changes
+       @side-by-side-active? ; re-render when side-by-side mode is toggled
+       (js/setTimeout
+        #(do
+           (when @map-a (.invalidateSize @map-a))
+           (when @map-b (.invalidateSize @map-b)))
+        50)))
```
markhepburn: That voodoo probably deserves a comment!

frontend/src/cljs/imas_seamap/natural_hazards_atlas/core.cljs
```diff
@@ -62,6 +68,7 @@
     :map/viewport-only?                   msubs/viewport-only?
     :sm/featured-maps                     smsubs/featured-maps
     :sm/featured-map                      smsubs/featured-map
+    :sok/boundary-layer-filter            (fn [] #(identity nil))
```
markhepburn: I assume that's a no-op hack?  Probalby worth at least commenting

frontend/src/cljs/imas_seamap/natural_hazards_atlas/utils.cljs
```diff
@@ -208,6 +208,20 @@
           (assoc m layer displayed-layer)))
       {}))))
 
+;; Proof-of-concept for having separate information in map B
+(defn layer-displayed-layers-lookup-map-b
```
markhepburn, line 262: This definitely needs a docstring when it moves out of proof-of-concept

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Taking a look at it now. I'll check if this is still necessary (and if it is, _why?_)

frontend/src/cljs/imas_seamap/natural_hazards_atlas/map/views.cljs
JoshGx2000: Removed the setTimeout. I had copied that from the (presumably Gemini-generated) code in the PoC:
https://github.com/condense/nhat-split-map-poc/blob/65bb1ca7c0bc9bac161053fecd0558e0f2ef4241/src/main/map_sync/views.cljs#L42-L44

It turned out to not be necessary though.

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Essentially, the problem is as follows:
- Panes are essential because they are used to control z-index of our layers, and mask layers in side-by-side layers (the feature from Seamap Antarctica).
- Panes require a name to function. If they aren't given a name, layers that are children of the pane are lumped in the default `leaflet-tile-pane`. 
- Panes need to have a unique name that doesn't clash with an existing pane's name. In theory `"<layer_id>"` is an acceptable name, since layers have entirely unique IDs. For side-by-side layers, `"<layer_id>-left"` and `"<layer_id>-right"` would be acceptable. 
- Panes don't appear to _ever_ be fully cleaned up. If (when) the map view re-renders, panes are re-created, but the old ones still exist on the map object. This causes a clash on the unique pane names, where the newly created panes error in their creation, insisting there is a pane that already exists with the unique name. 

We sidestep this issue currently by ensuring that every pane has a unique name every time it has to be recreated, but as a result we have this code smell.

Possible ways around this issue:
- "Figuring out" how to have the re-render recognise the leaflet panes aren't new panes but instead map to existing ones. Adding key props to the pane elements doesn't appear to assist with this.
- Clean-up of old panes between renders. Leaflet doesn't offer a way to remove panes without employing workarounds: https://github.com/Leaflet/Leaflet/issues/6408

That's my diagnosis so far @markhepburn. I do agree that the code is bad, but I've never come up with a neater solution than having unique pane names on each re-render.

frontend/src/cljs/imas_seamap/natural_hazards_atlas/core.cljs
JoshGx2000: Done

frontend/src/cljs/imas_seamap/natural_hazards_atlas/utils.cljs
JoshGx2000: Agreed. I've thought quite a bit about how separate information will work in a separate map. It'd be nice if we didn't have to have a whole new set of subs/events for the new map, but we also have the requirement of the existing subs/events working the same way they currently do, to maintain consistency with the other apps.

I have an idea on how it will work. I'll do a write-up for you to see if you agree with the approach, before beginning development.

frontend/src/cljs/imas_seamap/map/views.cljs
JoshGx2000: Just thought of it as I was leaving, but when I come in tomorrow I'll check if the `(.now js/Date)` part is necessary, or if `(random-uuid)` satisfies all requirements.

JoshGx2000: @markhepburn, it looks like `(str (random-uuid))` suffices for the condition of unique pane names withotu clashes. I've tested with re-ordering of layers and with the split-layers.

Do you have any thoughts on the matter, given the information I've shared so far?

frontend/src/cljs/imas_seamap/natural_hazards_atlas/utils.cljs
JoshGx2000: @markhepburn This is the rough idea for how I thought independent map state could be handled: https://docs.google.com/document/d/1zjd4pilrhll7KYKhnYORKwLRFlxLM9EWAjZD8yh3BOo/edit?usp=sharing

Not finished writing the document yet.

I think the approach I've written is not _entirely_ needed, as how independent the two map states will be has been scaled-back a lot.

Do let me know of your rough opinion on the matter.

